PR #50, opened 2018-04-22
add crashlytics to sample app

YohayBar:


app/build.gradle
```diff
@@ -1,4 +1,26 @@
...
+        maven { url 'https://maven.fabric.io/public' }
+    }
+
+    dependencies {
+        // These docs use an open ended version so that our plugin
+        // can be updated quickly in response to Android tooling updates
+
+        // We recommend changing it to the latest version from our changelog:
+        // https://docs.fabric.io/android/changelog.html#fabric-gradle-plugin
+        classpath 'io.fabric.tools:gradle:1.+'
```
OrenZak: you can add it directly to our dependencies, there is no need for `buildscript` closure.
also, you can change the fabric tools version to: `1.25.3` instead of having 1.+
and remove the top maven source you have it below

OrenZak: replace the version with : `1.25.3`

app/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,10 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <meta-data
+            android:name="io.fabric.ApiKey"
+            android:value="e5627bac828b2b0b56a64f44f0ffba67f68324fa"
```
OrenZak: in that case, whoever will run the sample and make changes that will cause a crash also we will be notified.

would be better to be something in ignore right?

PR #67, opened 2018-05-17
fix redeem code pops up twice

OrenZak:
the observer wasn't removed so we got onChange twice,
also noticed that get order call wasn't getting a delay between each call, because it was milliseconds


kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/data/order/OrderRepository.java
```diff
@@ -152,23 +150,21 @@ private void getOrder(final String orderID) {
...
             }
         });
     }
 
     private void decrementPendingOrdersCount() {
         if (hasMorePendingOffers()) {
-            pendingOrdersCount.decrementAndGet();
-        } else {
-            if (paymentObserver != null) {
+            if (pendingOrdersCount.decrementAndGet() == 0 && paymentObserver != null) {
```
YohayBar: as we discussed add ~ paymentObserverCount to only control adding/removing  payment observer

OrenZak: 👍

PR #77, opened 2018-05-28
add prefix to all res values

OrenZak:
We could have a conflict with the hosting app, so now we got a prefix for all the res values to resolve that issue


YohayBar:
see also https://developer.android.com/studio/projects/android-library on how to make the resource private

PR #98, opened 2018-06-25
Add closeMarketplace functionality to the sdk.

Blahartinger:
This exposes a public ```closeMarketplace``` function on the Kin class to allow callers of
```launchMarketplace ``` a companion method to programatically close it.
e.g. This is useful in cases such as a NativeSpendOffer registered Observer callback.

*Technical Description*
Utilizes Local Broadcast messages and BroadcastReciever's on both the Marcketplace and SplashView activities.


OrenZak:
We are going to implement it in another way.
Thanks for the help!

PR #99, opened 2018-06-25
bi events fixes

OrenZak:


kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/data/auth/AuthLocalData.java
```diff
@@ -120,15 +122,21 @@ public String getUserID() {
 		return signInSharedPreferences.getString(USER_ID_KEY, null);
 	}
 
+	@Override
+	public String getEcosystemUserID() {
+		return signInSharedPreferences.getString(ECOSYSTEM_USER_ID_KEY, null);
```
YohayBar: default string  "N/A" or "null" maybe?

kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/data/auth/AuthLocalData.java
```diff
@@ -120,15 +122,21 @@ public String getUserID() {
...
+	}
+
 	@Override
 	public AuthToken getAuthTokenSync() {
 		String token = signInSharedPreferences.getString(TOKEN_KEY, null);
 		String appID = signInSharedPreferences.getString(APP_ID_KEY, null);
 		String userID = signInSharedPreferences.getString(USER_ID_KEY, null);
+		String ecosystemUserID = signInSharedPreferences.getString(ECOSYSTEM_USER_ID_KEY, null);
```
YohayBar: use getEcosystemUserID()

kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/data/auth/AuthLocalData.java
OrenZak: for the local should return as null not empty string or something

kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/data/auth/AuthLocalData.java
OrenZak: just keeping the same for every param... 
we can change but doesn't really matter

kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/EventCommonDataUtil.java
```diff
@@ -113,4 +114,8 @@ public String get() {
 		EventsStore.update(commonModifier);
 		EventsStore.update(clientModifier);
 	}
+
+	private static String checkNullString(final String text) {
+		return text != null ? text : "null";
```
YohayBar: safeguardNullString(final String text) maybe?

kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/data/auth/AuthLocalData.java
YohayBar: just duplicate code - but not critical

kin-ecosystem-sdk/src/main/java/com/kin/ecosystem/poll/presenter/PollWebViewPresenter.java
```diff
@@ -70,7 +71,7 @@ private void setTitle(String title) {
 	}
 
 	private void createOrder() {
-		EarnOrderCreationRequested.create(EarnOrderCreationRequested.OfferType.fromValue(contentType), (double) amount, offerID);
+		eventLogger.send(EarnOrderCreationRequested.create(EarnOrderCreationRequested.OfferType.fromValue(contentType), (double) amount, offerID));
```
YohayBar: EarnOrderCreationRequested.OfferType.fromValue(contentType) can throw runtime exception for example if we add new contentType in the future old SDK will crash because  OfferType.fromValue throws IllegalArgumentException if it receive a contentType not know - I think it safe to just pass the   contentType itself - worst case scenario we are in validation of the json schema but the event will be sent. This can be true to any other place we are using fromValue();

PR #127, opened 2018-08-05
Change order after transaction failed with the Error

OrenZak:
#### Main purpose:
Change the order if happened any failure on the client side that the server should know about.
For example, if a `sendTransaction` -> failed so we can notify the server about it and he will stop listening for the payment to be received on the blockchain side. So the user/SDK developer will get a failure faster.
#### Technical description:
- Added `changeOrder` API on `OrderRemoteData.java`
- Added login on `OrderRepository.java` so when payment is failed we call `changeOrder` with Transaction failed `Error`.
- Removed the cancel call on `onTransactionFailed` callback from purchase call, because as @doodyparizada  said we can't cancel an order if it already created.
#### Tests:
- Changed `purchase` test to verify we are calling to `remote.changeOrder` when the payment is failed, and never call it if the payment is succeeded.
#### Documentation:
- Added on `OrderRepository.java` for `changeOrder` method.


core/src/main/java/com/kin/ecosystem/core/data/order/OrderRemoteData.java
```diff
@@ -243,6 +244,50 @@ public void getFilteredOrderHistory(@Nullable String origin, @NonNull String off
...
+                    executorsUtil.mainThread().execute(new Runnable() {
+                        @Override
+                        public void run() {
+                            callback.onResponse(result);
+                        }
+                    });
+				}
+
+				@Override
+				public void onUploadProgress(long bytesWritten, long contentLength, boolean done) {
```
yosriz: why are those callbacks needed? what do we upload here? I understand it's auto generated from the swagger definition, but still wondering

core/src/main/java/com/kin/ecosystem/core/network/model/Body.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+
+import com.google.gson.annotations.SerializedName;
+import java.util.Objects;
+
+/**
+ * Body
+ */
+public class Body {
```
yosriz: what's the purpose of this `Body` class? it seems like a wrapper for error, we can't we use Error directly?

core/src/test/java/com/kin/ecosystem/core/data/order/OrderRepositoryTest.java
```diff
@@ -548,6 +550,7 @@ public void onFailure(KinEcosystemException exception) {
 
 		verify(blockchainSource).removePaymentObserver(observersList.get(0));
 		verify(blockchainSource).removePaymentObserver(observersList.get(1));
+		verify(remote).changeOrder(anyString(), any(Body.class), any(Callback.class));
```
yosriz: can't you verify the actual input? making sure they match the expected ones?

core/src/main/java/com/kin/ecosystem/core/data/order/OrderRemoteData.java
OrenZak: no need, generated swagger

core/src/test/java/com/kin/ecosystem/core/data/order/OrderRepositoryTest.java
OrenZak: nope, the `Body` and `Callback` created inside the repo.

core/src/main/java/com/kin/ecosystem/core/network/model/Body.java
OrenZak: We can use, but for future maybe we will want to add something extra without affecting the Error class. Also this is something generated from swagger and i don't wanna change that

PR #129, opened 2018-08-06
Fail on requests without auth token

OrenZak:
#### Main purpose:
Do not call to the server without `AuthToken` so we can reduce request count.
#### Technical description:
Added logic on the interceptor of our network calls, if there is no `AuthToken` on response with a failure from the client and don't send the req to the server.


core/src/main/java/com/kin/ecosystem/core/Configuration.java
```diff
@@ -57,20 +63,29 @@ public static ApiClient getDefaultApiClient() {
...
-							Request authorisedRequest = originalRequest.newBuilder()
-								.header(AUTHORIZATION, BEARER + authToken.getToken())
-								.build();
-							return chain.proceed(authorisedRequest);
-						} else {
+						if (path.equals(USERS_PATH)) {
 							return chain.proceed(originalRequest);
+						} else {
+							AuthToken authToken = AuthRepository.getInstance().getAuthTokenSync();
+							if (authToken != null) {
```
YohayBar: make sure to request offer in Kin after register for the first time to pre fetch offers

sdk/src/main/java/com/kin/ecosystem/splash/presenter/SplashPresenter.java
```diff
@@ -123,6 +127,7 @@ public void run() {
 				stopLoading(true);
 				showTryAgainLater();
 				removeAccountStateObserver();
+				shouldShowError = false;
```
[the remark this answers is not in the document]
YohayBar: this fix the showing error twice bug?

core/src/main/java/com/kin/ecosystem/core/Configuration.java
```diff
@@ -57,20 +63,29 @@ public static ApiClient getDefaultApiClient() {
...
+								Request authorisedRequest = originalRequest.newBuilder()
+									.header(AUTHORIZATION, BEARER + authToken.getToken())
+									.build();
+								return chain.proceed(authorisedRequest);
+							} else {
+								// Stop the request from being executed.
+								Logger.log(new Log().withTag("ApiClient").text("No token - response error on client"));
+								return new Response.Builder()
+									.code(NO_TOKEN_ERROR_CODE)
```
YohayBar: you can move the error response response builded to a method for better readability (IMO)

PR #134, opened 2018-08-12
Call submitOrder before send transaction to blockchain

OrenZak:
#### Main purpose:
We need to call submitOrder before sending the transaction to the blockchain, so if the transaction is sent before submitOrder, the server might miss the transaction
#### Technical description:
- Moved `sendTransaction` after submitOrder is completed.


core/src/main/java/com/kin/ecosystem/core/data/order/ExternalSpendOrderCall.java
```diff
@@ -3,16 +3,15 @@
...
 
 class ExternalSpendOrderCall extends CreateExternalOrderCall {
 
 	ExternalSpendOrderCall(
-		@NonNull Remote remote,
+		@NonNull OrderDataSource orderRepository,
 		@NonNull BlockchainSource blockchainSource,
 		@NonNull String orderJwt,
 		@NonNull EventLogger eventLogger,
-		@NonNull ExternalSpendOrderCallbacks externalSpendOrderCallbacks) {
```
OrenZak: revert callbacks to old one

PR #136, opened 2018-08-26
Eco 539/test to account manager with kotlin

OrenZak:
#### Main purpose:
Validate that each function performs as designed
#### Technical description:
- Added unit test for AccountManager
- Added Kotlin for tests
#### Dilemmas you faced with? 
which Kotlin testing lib to use, found that Mockito teams say that there is a Kotlin version of their lib.
I gave it a test :) and it works really good and has great options to use.
So it's basically mockito2 in Kotlin.
Maintained by Niek Haarman, see the [Github](https://github.com/nhaarman/mockito-kotlin) project.
#### Tests:
- get account status
- start full account creation flow
- retry for a full account creation flow when the state is Error


core/src/test/kotlin/com/kin/ecosystem/core/accountmanager/AccountManagerImplTest.kt
```diff
@@ -0,0 +1,152 @@
...
+
+        accountManager.addAccountStateObserver(stateObserver)
+        accountManager.start()
+        argumentCaptor<EventListener<Void>>().apply {
+            verify(blockchainEvents).addAccountCreationListener(capture())
+            firstValue.onEvent(null)
+        }
+
+
+
```
nitzantomer: I'd add another empty line or two, just to make a point.

core/src/test/kotlin/com/kin/ecosystem/core/accountmanager/AccountManagerImplTest.kt
```diff
@@ -0,0 +1,152 @@
...
+        assertEquals(listOf(1, 1, 2, 3, 4), stateList)
+    }
+
+
```
nitzantomer: What, you're not following a set of conventions? Why two empty lines here, but above you only had one?

test-base/src/main/java/kin/ecosystem/test/base/BaseTestClass.kt
```diff
@@ -0,0 +1,22 @@
+package kin.ecosystem.test.base
+
+import com.kin.ecosystem.core.bi.EventsStore
+import org.junit.Before
+
+
+open class BaseTestClass {
+
+    protected var userModifierFake = UserModifierFake()
```
nitzantomer: why are these three aren't `val`s?

test-base/src/main/java/kin/ecosystem/test/base/ClientModifierFake.kt
```diff
@@ -0,0 +1,33 @@
+package kin.ecosystem.test.base
+
+import com.kin.ecosystem.core.bi.EventsStore.ClientModifier
+import com.kin.ecosystem.core.bi.EventsStore.DynamicValue
+import com.kin.ecosystem.core.bi.events.ClientProxy
+
+class ClientModifierFake : ClientModifier {
+
+    private var os: String? = null
```
nitzantomer: maybe use `lateinit` instead of assigning `null`s?
or maybe you don't even need to because you're using the `init` block?

test-base/src/main/java/kin/ecosystem/test/base/CommonModifierFake.kt
```diff
@@ -0,0 +1,28 @@
+package kin.ecosystem.test.base
+
+import com.kin.ecosystem.core.bi.EventsStore.CommonModifier
+import com.kin.ecosystem.core.bi.EventsStore.DynamicValue
+import com.kin.ecosystem.core.bi.events.CommonProxy
+import java.util.UUID
+
+class CommonModifierFake : CommonModifier {
+
+    private var eventId: UUID? = null
```
nitzantomer: same comment as in `ClientModifierFake.kt`

test-base/src/main/java/kin/ecosystem/test/base/UserModifierFake.kt
```diff
@@ -0,0 +1,42 @@
+package kin.ecosystem.test.base
+
+import com.kin.ecosystem.core.bi.EventsStore.DynamicValue
+import com.kin.ecosystem.core.bi.EventsStore.UserModifier
+import com.kin.ecosystem.core.bi.events.UserProxy
+
+class UserModifierFake : UserModifier {
+
+    private var digitalServiceUserId: String? = null
```
nitzantomer: same as the last two comments

nitzantomer:
i wrote some comments, i don't think that anything there is a must, just some things that looked weird or might needed to change.
i'll approve it so that it won't get stuck while we're all abroad.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -324,6 +328,33 @@ public void removePaymentObserver(Observer<Payment> observer) {
 		decrementPaymentCount();
 	}
 
+	@Override
+	public void createTrustLine(@NonNull final KinCallback<Void> callback) {
+		new CreateTrustLineCall(account, new TrustlineCallback() {
```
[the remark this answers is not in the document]
OrenZak: i'll open a task

test-base/src/main/java/kin/ecosystem/test/base/ClientModifierFake.kt
OrenZak: it was auto conversion from Java to Kotlin by the IDE

PR #137, opened 2018-08-26
BI - spend order completed fired twice

OrenZak:
#### Main purpose:
BI - event `spend_order_completed` was fired twice because we didn't have the origin of the order. 
I found another bug that we have sent also `earn_order_payment_confirmed` on each payment confirmation and only on earn orders.
#### Technical description:
- Check whether the order origin is MARKETPLACE on `sendSpendOrderCompleted`.
The Native order (Origin: EXTERNAL) is called on its own logic.
- Added to `Payment.java` type filed, and check if this field is EARN on `sendEarnPaymentConfirmed`
#### Tests:
- updated order origin MARKETPLACE where needed
- updated payment type (EARN / SPEND/UNKNOWN)
#### Documentation:
- Added on the source for `Payment -> type`

### DIFF
--- a/core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
+++ b/core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
@@ -6,18 +6,20 @@ import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
 import android.text.TextUtils;
 import com.kin.ecosystem.common.KinCallback;
-import com.kin.ecosystem.common.exception.BlockchainException;
-import com.kin.ecosystem.core.Log;
-import com.kin.ecosystem.core.bi.events.KinBalanceUpdated;
-import com.kin.ecosystem.core.Logger;
+import com.kin.ecosystem.common.KinCallbackAdapter;
 import com.kin.ecosystem.common.ObservableData;
 import com.kin.ecosystem.common.Observer;
-import com.kin.ecosystem.common.KinCallbackAdapter;
+import com.kin.ecosystem.common.exception.BlockchainException;
+import com.kin.ecosystem.common.model.Balance;
+import com.kin.ecosystem.core.Log;
+import com.kin.ecosystem.core.Logger;
 import com.kin.ecosystem.core.bi.EventLogger;
+import com.kin.ecosystem.core.bi.events.KinBalanceUpdated;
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -311,6 +313,11 @@ public void onResponse(Balance response) {
 			});
 	}
 
+	private Payment toPaymentObj(final String orderID, PaymentInfo data) {
```
YohayBar: toPaymentObj should be in Payment.java - used twice every future update will need to be updated in several places

core/src/main/java/com/kin/ecosystem/core/data/blockchain/Payment.java
```diff
@@ -9,69 +12,88 @@
...
-        return exception;
-    }
-
-	public boolean isEarn() {
-		return amount != null && amount.compareTo(BigDecimal.ZERO) == 1;
+	public static final int UNKNOWN = 0x00000000;
+	public static final int EARN = 0x00000001;
+	public static final int SPEND = 0x00000002;
+
+	@IntDef({EARN, SPEND})
```
YohayBar: why UNKNOWN is not part of the @IntDef

core/src/main/java/com/kin/ecosystem/core/data/blockchain/Payment.java
```diff
@@ -9,69 +12,88 @@
...
+		this.isSucceed = true;
+		this.type = type;
+	}
+
+	public Payment(String orderID, boolean isSucceed, Exception error) {
+		this.orderID = orderID;
+		this.transactionID = null;
+		this.amount = null;
+		this.isSucceed = isSucceed;
+		this.exception = error;
```
YohayBar: error doesn't need type?

OrenZak: type default is UNKNOWN

core/src/main/java/com/kin/ecosystem/core/data/blockchain/Payment.java
OrenZak: forgot to add.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: it used only once.

PR #138, opened 2018-08-27
bi earn/spend order cration faild blank error msg

OrenZak:
#### Main purpose:
BI - event `earn_order_creation_failed` was fired with blank error message sometimes.
#### Technical description:
- Send the error message from the server.

### DIFF
--- a/sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/SpendDialogPresenter.java
+++ b/sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/SpendDialogPresenter.java
@@ -19,6 +19,7 @@ import com.kin.ecosystem.common.KinCallbackAdapter;
 import com.kin.ecosystem.core.data.blockchain.BlockchainSource;
 import com.kin.ecosystem.core.data.order.OrderDataSource;
 import com.kin.ecosystem.common.exception.KinEcosystemException;
+import com.kin.ecosystem.core.network.ApiException;
 import com.kin.ecosystem.marketplace.view.ISpendDialog;
 import com.kin.ecosystem.core.network.model.Offer;
 import com.kin.ecosystem.core.network.model.OfferInfo;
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/SpendDialogPresenter.java
```diff
@@ -83,8 +84,12 @@ public void onResponse(OpenOrder response) {
 			@Override
 			public void onFailure(KinEcosystemException exception) {
 				showToast("Oops something went wrong...");
-				eventLogger
-					.send(SpendOrderCreationFailed.create(exception.getCause().getMessage(), offer.getId(), false));
+				try {
+					String errorMsg = ((ApiException) exception.getCause()).getResponseBody().getMessage();
+					eventLogger.send(SpendOrderCreationFailed.create(errorMsg, offer.getId(), false));
+				} catch (ClassCastException e) {
+					eventLogger.send(SpendOrderCreationFailed.create(exception.getMessage(), offer.getId(), false));
```
YohayBar: it will be more readable (IMO) if try catch will only affect String errorMsg and after the try catch you will call 'eventLogger.send(SpendOrderCreationFailed.create(errorMsg, offer.getId(), false));'  once.

sdk/src/main/java/com/kin/ecosystem/poll/presenter/PollWebViewPresenter.java
```diff
@@ -86,9 +87,12 @@ public void onResponse(OpenOrder response) {
...
 			public void onFailure(KinEcosystemException exception) {
-				eventLogger.send(EarnOrderCreationFailed.create(exception.getCause().getMessage(), offerID));
-				if (view != null) {
-					showToast(exception.getMessage());
+				showToast("Oops something went wrong...");
+				try {
+					String errorMsg = ((ApiException) exception.getCause()).getResponseBody().getMessage();
+					eventLogger.send(EarnOrderCreationFailed.create(errorMsg, offerID));
+				} catch (ClassCastException e) {
+					eventLogger.send(EarnOrderCreationFailed.create(exception.getMessage(), offerID));
```
YohayBar: same try catch should only affect String errorMsg

sdk/src/main/java/com/kin/ecosystem/poll/presenter/PollWebViewPresenter.java
```diff
@@ -86,9 +87,12 @@ public void onResponse(OpenOrder response) {
 
 			@Override
 			public void onFailure(KinEcosystemException exception) {
-				eventLogger.send(EarnOrderCreationFailed.create(exception.getCause().getMessage(), offerID));
-				if (view != null) {
-					showToast(exception.getMessage());
+				showToast("Oops something went wrong...");
```
YohayBar: 1. string should be in strings.xml 
2. Why not displaying the error message in the toast (it will help us debug the problem)

sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/SpendDialogPresenter.java
OrenZak: 👍

PR #139, opened 2018-08-27
BI - add proprty to spend order completed

OrenZak:
#### Main purpose:
Add new poperty `kin_amount` to `spen_order_completed`
#### Technical description:
Copy generated file, and add the value where used.


YohayBar:
fix conflict

PR #140, opened 2018-08-29
BI - kin sdk initiated missing value

OrenZak:
#### Main purpose:
BI event `kin_sdk_initiated` property `digital_service_id` was always null value
#### Technical description:
Changed `getAppId`  on `Local` side to be sync and not async. (it's from the shared_pref) so it's ok.
Making sure the value is set, saved and used


core/src/main/java/com/kin/ecosystem/core/data/auth/AuthLocalData.java
```diff
@@ -92,24 +92,8 @@ public void run() {
...
-						} else {
-							callback.onFailure(null);
-						}
-					}
-				});
-			}
-		};
-		executorsUtil.diskIO().execute(command);
+	public String getAppId() {
+		return signInSharedPreferences.getString(APP_ID_KEY, null);
```
YohayBar: need to also change how raw APP_ID_KEY is saved currently it commit with IO thread - it should be just applied on UI thread (need to change it 2 methods)

PR #144, opened 2018-09-03
API - hasAccount: determine if a Kin Account is associated with a userId

OrenZak:
#### Main purpose:
While we developed the option to send Kin from peer to peer, it is necessary to have the ability to determine if a Kin Account is associated with some userId before we send the Kin.
#### Technical description:
- Add new API call `hasAccount(userId, callback)` to `Kin.java`
- Update Sample app so it will check if a userId has an account before paying to that user.
#### Documentation:
- Source code
- README -> /docs/Misc.md


app/src/main/java/com/ecosystem/kin/app/MainActivity.java
```diff
@@ -346,6 +333,46 @@ private void createPayToUserOffer(String recipientUserID) {
 		}
 	}
 
+	private void showPayToUserDialog(final View v) {
+		PayToUserDialog dialog = new PayToUserDialog(v.getContext());
+		dialog.setOnDismissListener(new OnDismissListener() {
+			@Override
+			public void onDismiss(DialogInterface dialog) {
```
nitzantomer: it is confusing that you named this variable `dialog` when you already have one variable with that name

app/src/main/java/com/ecosystem/kin/app/MainActivity.java
```diff
@@ -346,6 +333,46 @@ private void createPayToUserOffer(String recipientUserID) {
...
+				PayToUserDialog myDialog = (PayToUserDialog) dialog;
```
nitzantomer: this is redundant.  just mark the top `dialog` as `final` and use it

PR #146, opened 2018-09-04
BI - update spend events with origin field

OrenZak:
#### Main purpose:
We are going to replace `is_native` field with `origin` for consistency.
First, add the `origin` field and later on, we will remove `is_native`.
#### Technical description:
- Copy generated files
- Add `origin` value where is needed


OrenZak:
They should yup :)

PR #150, opened 2018-09-13
fix accepted language format

YohayBar:
#### Main purpose:
fix accepted language format.

### DIFF
--- a/core/src/main/java/com/kin/ecosystem/core/Configuration.java
+++ b/core/src/main/java/com/kin/ecosystem/core/Configuration.java
@@ -2,6 +2,7 @@ package com.kin.ecosystem.core;
 
 import android.os.Build;
 import android.os.Build.VERSION;
+import android.text.TextUtils;
 import com.kin.ecosystem.common.KinEnvironment;
 import com.kin.ecosystem.core.data.auth.AuthRepository;
 import com.kin.ecosystem.core.network.ApiClient;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/kin/ecosystem/core/Configuration.java
```diff
@@ -104,4 +105,15 @@ public static void setEnvironment(KinEnvironment environment) {
 
 	private Configuration() {
 	}
+
+	private static String getDeviceAcceptedLanguage() {
+		Locale defaultLocale = Locale.getDefault();
+		String acceptedLanguage;
```
nitzantomer:
```
String acceptedLanguage = defaultLocale.getLanguage();
if (!TextUtils.isEmpty(defaultLocale.getCountry())) {
	acceptedLanguage += "-" + defaultLocale.getCountry();
}
```

core/src/main/java/com/kin/ecosystem/core/Configuration.java
```diff
@@ -104,4 +105,15 @@ public static void setEnvironment(KinEnvironment environment) {
...
+		if (TextUtils.isEmpty(defaultLocale.getCountry())) {
+			acceptedLanguage = defaultLocale.getLanguage();
```
liorama: @YohayBar are we sure `defaultLocale.getLanguage()` always returns a 2 character code?

PR #153, opened 2018-10-03
Move every hard coded string to strings.xml

OrenZak:
#### Main purpose:
More language support
#### Technical description:
Move every hard coded string to strings.xml


sdk/src/main/java/com/kin/ecosystem/history/view/OrderHistoryRecyclerAdapter.java
```diff
@@ -65,6 +65,12 @@ private void initSizes(Context context) {
         }
     }
 
+    private void initStrings(Context context) {
+        if(TRANSACTION_FAILED_MSG == null) {
+            TRANSACTION_FAILED_MSG = context.getString(R.string.kinecosystem_transaction_failed);
```
YohayBar: I don't like TRANSACTION_FAILED_MSG being null. I understand that init(context) must be called first otherwise we don't have context but having null open a window (very small one I admit) for NPE

sdk/src/main/java/com/kin/ecosystem/splash/view/SplashActivity.java
```diff
@@ -28,153 +30,155 @@
...
+		});
+	}
+
+	@NonNull
+	private @StringRes int getMessageResId(@Message final int msg) {
+		switch (msg) {
+			case TRY_AGAIN:
+				return R.string.kinecosystem_try_again_later;
+			default:
+			case SOMETHING_WENT_WRONG:
```
YohayBar: what was changed in this file?

OrenZak: only showToast and getMessageResId.
maybe the rest related to indentation.

sdk/src/main/java/com/kin/ecosystem/history/view/OrderHistoryRecyclerAdapter.java
OrenZak: i'll set empty value

PR #156, opened 2018-10-11
Add KeyStoreProvider Implementation

OrenZak:
#### Main purpose:
Add KeyStoreProvider Implementation so we can start work on backup and restore flows.
#### Technical description:
- Add KeyStoreProvider Implementation
- Add Backup exceptions
#### Tests:
- BlockchainSource getKeyStoreProvider is not null.
#### Documentation:
- Added to source

### DIFF
--- a/core/src/main/java/com/kin/ecosystem/core/data/blockchain/KeyStoreProviderImpl.java
+++ b/core/src/main/java/com/kin/ecosystem/core/data/blockchain/KeyStoreProviderImpl.java
@@ -23,7 +23,7 @@ public class KeyStoreProviderImpl implements KeyStoreProvider {
 	private final KinClient kinClient;
 	@Nullable
 	private final KinAccount kinAccount;
-	private Pattern pattern;
+	private final Pattern pattern;
 
 	KeyStoreProviderImpl(@NonNull final KinClient kinClient, @NonNull final KinAccount kinAccount) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/kin/ecosystem/core/data/blockchain/KeyStoreProviderImpl.java
```diff
@@ -1,22 +1,80 @@
...
+		} catch (CorruptedDataException e) {
+			throw new BackupException(CODE_RESTORE_INVALID_KEYSTORE_FORMAT, "The keystore is invalid - wrong format");
+		}
+	}
+
+	@Override
+	public boolean validatePassword(@NonNull final String password) {
+		Validator.checkNotNull(password, "password");
+		Pattern pattern = Pattern
+			.compile("^(?=.*\\d)(?=.*[A-Z])(?=.*[!@#$%^&*()_+{}\\[\\]])(?!.*[^a-zA-Z0-9!@#$%^&*()_+{}\\[\\]])(.{9,})$");
```
yosriz: reuse Pattern (create it once), Regex pattern compile might be time consuming and affects UI (as it will user directly on each keystroke on the UI)

OrenZak: 👍

PR #157, opened 2018-10-15
Add QR lib to use in backup and restore

yosriz:
#### Main purpose:
add QR decoding/encoding + saving/loading qr image file utilities
#### Dilemmas you faced with? 
QR - choosed zxing as it's the standard, popular lib, google now have QR scanning in ml-vision, but that's only one sided (decoding).
I took only core cause we don't need camera scanning, very little additional code required for preparing the input for zxing.
Image saving/loading - using fileprovider to save it in private app data and to provide it as URI only when we share the QR

#### Tests:
tests for QR decoding added (with real expected keystore data)
#### Documentation:
N/A


backup/src/main/java/com/kin/ecosystem/backup/qr/QRFileUriHandlerImpl.java
```diff
@@ -0,0 +1,81 @@
...
+		}
+		throw new IOException("decoding file as bitmap failed.");
+	}
+
+	@NonNull
+	@Override
+	public Uri saveFile(@NonNull Bitmap bitmap) throws IOException {
+		File file = getOrCreateSaveFile();
+		FileOutputStream stream = new FileOutputStream(file);
+		bitmap.compress(Bitmap.CompressFormat.PNG, 90, stream);
```
YohayBar: do we ever delete this file?

yosriz: we can't do it reliably, meaning we don't know when the file can be deleted.

PR #169, opened 2018-10-27
Eco 668/switch account

yosriz:
#### Main purpose:
Switching account after restore
#### Technical description:
Switching account steps:
 - Switch to the new `KinAccount` at `BlockchainSourceImpl` - will persist
  current account index and update the active account
 - Update persisted sign in data with new wallet address at `AuthRepository`
 - Force `REQUIRE_CREATION` account state, in order to trigger re-sign to
  server for updating the new wallet address
#### Dilemmas you faced with? 
* TODO - need to add logging/BI event after a failure of account switching (at setting presenter)
* Currently only the happy path is covered, if any failure with switching happened we'll end up in an unexpected state. maybe a blocking UI is needed when switching?
* hardcoded `schemaVersion` to "1" at `CommonProxy` to prevent crash and tests failure
* rebased on top of "Eco 647 restore flow", only last commit relevant
#### Tests:
N/A - tested manually
#### Documentation:
N/A


app/src/main/java/com/ecosystem/kin/app/App.java
```diff
@@ -22,7 +22,7 @@ public void onCreate() {
 
         Fabric.with(this, new Crashlytics());
 
-		KinEnvironment environment = Environment.getPlayground();
+		KinEnvironment environment = Environment.getBeta();
```
YohayBar: change to playground in dev?

core/src/main/java/com/kin/ecosystem/core/accountmanager/AccountManagerImpl.java
```diff
@@ -169,14 +168,35 @@ public void onFailure(KinEcosystemException error) {
...
+		final String address = blockchainSource.getPublicAddress(accountIndex);
+		//update sign in data with new wallet address and update servers
+		authRepository.updateWalletAddress(address, new KinCallback<Boolean>() {
+			@Override
+			public void onResponse(Boolean response) {
+				try {
+					//switch to the new KinAccount
+					blockchainSource.updateActiveAccount(accountIndex);
+				} catch (BlockchainException e) {
+					callback.onFailure(ErrorUtil.getBlockchainException(e));
```
YohayBar: Why blockchain exception it will be a server exception

YohayBar: add logger to catch?

recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenterImpl.java
```diff
@@ -1,37 +1,66 @@
...
+				}
+				step--;
+				view.onBackButtonClicked();
+			}
 		}
 	}
 
 	@Override
-	public void setStep(@Step final int step) {
+	public void setStep(@Step final int step, @Nullable Bundle data) {
```
YohayBar: instead of calling setStep with null you can add public void setStep(@Step final int step) that will call the second one with null

sdk/src/main/java/com/kin/ecosystem/Environment.java
```diff
@@ -75,4 +83,8 @@ public static KinEnvironment getProduction() {
 	public static KinEnvironment getPlayground() {
 		return PLAYGROUND;
 	}
+
+	public static KinEnvironment getBeta() {
+		return BETA;
```
YohayBar: add documentation that Beta is for internal testing. Developers should use playground for testing and Production for release app

sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.java
```diff
@@ -174,6 +170,26 @@ public void onFailure(Throwable throwable) {
 		});
 	}
 
+	private void switchAccount(int accountIndex) {
+		accountManager.switchAccount(accountIndex, new KinCallback<Boolean>() {
+			@Override
+			public void onResponse(Boolean response) {
+				//do nothing succeed
```
YohayBar: Ask Noa - (not now) if we need to show toast message when returning to setting?

PR #176, opened 2018-11-02
Avoid conflicts with other SDKs that are using FileProvider.

phips28:
Fixes: https://github.com/kinecosystem/kin-ecosystem-android-sdk/issues/175

Found solution here: https://github.com/stkent/bugshaker-android/issues/108#issuecomment-293291649

PR #177, opened 2018-11-04
Create uniqe content file provider authority

OrenZak:
#### Main purpose:
Allow using the SDK in multiple apps, and create unique authorities
#### Technical description:
setting the authorities start with applicationId from the host application.
#### Tests:
Install a few apps with the SDK and check that there is no conflicts.


recovery/src/main/AndroidManifest.xml
```diff
@@ -3,14 +3,14 @@
 
     <application>
         <provider
-            android:authorities="com.kin.ecosystem.recovery"
+            android:authorities="${applicationId}.KinRecoveryFileProvider"
             android:exported="false"
             android:grantUriPermissions="true"
             android:name="android.support.v4.content.FileProvider">
```
YohayBar: In the end, you didn't change android:name - I don't think we need to just to make sure it on purpose?

phips28:
Keep that in mind :)
https://github.com/kinecosystem/kin-ecosystem-android-sdk/pull/176

yosriz:
@OrenZak additional comment, I think that @phips28 solution in https://github.com/kinecosystem/kin-ecosystem-android-sdk/pull/176 should be treated as well.
If I understand correctly, there are 2 issues here that we should address:
* A conflict with multiple apps that use our SDK, because of the same authority.
* A conflict with multiple providers of `FileProvider` in the *same* app, caused by the same provider name.

PR #178, opened 2018-11-05
Add screen size support

OrenZak:
#### Main purpose:
Fix some UI issues related to the different screen size.
#### Technical description:
create dimens values for different screen size.
#### Tests:
tested on most used screen size (hdpi, xhdpi,xxhdpi)


recovery/src/main/java/com/kin/ecosystem/recovery/widget/PasswordEditText.java
```diff
@@ -94,7 +95,7 @@ private void setupPasswordField(boolean addRevealIcon, String hint) {
 		passwordField.setTypeface(Typeface.SANS_SERIF);
 		passwordField.setTextColor(colorGray);
 		passwordField.setHintTextColor(colorGray);
-		passwordField.setTextSize(TypedValue.COMPLEX_UNIT_SP, 14f);
+		passwordField.setTextSize(TypedValue.COMPLEX_UNIT_PX, textSize);
```
yosriz: why changing to PX instead of SP?

OrenZak: because textsize comes as px from resources.

PR #182, opened 2018-11-06
Fix settings icon menu

OrenZak:
#### Main purpose:
Settings icon was huge red dot on some devices.


yosriz:
what's the purpose of using action in the menu item?

OrenZak:
@yosriz The Android OS acts differently for just drawables and for actionViews.
So when it's drawable is set it with a default size, so you can't control the drawable, therefore we saw a big red dot when the settings drawable contained the oval shape.
And using the actionView lets you the option to control the whole menu item UI.

yosriz:
@OrenZak :+1: thanks

yosriz (approved):

PR #184, opened 2018-11-07
Eco 738 - keyboard stays open after complete restore flow

OrenZak:
#### Main purpose:
Close keyboard after complete restore flow.

* Also noticed that keyboard does not open on EnterPassword fragment due to system delay on getting view boundaries, so I added a minimum delay (otherwise it won't open).


recovery/src/main/java/com/kin/ecosystem/recovery/restore/view/RestoreEnterPasswordFragment.java
```diff
@@ -116,7 +115,12 @@ public void afterTextChanged(Editable editable) {
 				presenter.onPasswordChanged(editable.toString());
 			}
 		}));
-		openKeyboard(password);
+		password.postDelayed(new Runnable() {
```
yosriz, line 118: do you know why the delay is needed? maybe an event can be used instead? (like `OnGlobalLayoutListener`)

OrenZak: i'll check

OrenZak: @yosriz I've tried, its not working.

yosriz (approved):

PR #185, opened 2018-11-07
Solved crash after activity is killed for backup flow

OrenZak:
#### Main purpose:
The crash appeared while having the developer option of `don't keep activities`.
This fix is only for backup flow, restore flow will do next.
#### Technical description:
Add save instance state mechanism for the fragments and main activity `BackupActivity`

### DIFF
--- a/recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenter.java
+++ b/recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenter.java
@@ -1,9 +1,13 @@
 package com.kin.ecosystem.recovery.backup.presenter;
 
+import android.os.Bundle;
 import com.kin.ecosystem.recovery.backup.view.BackupNextStepListener;
 import com.kin.ecosystem.recovery.backup.view.BackupView;
 import com.kin.ecosystem.recovery.base.BasePresenter;
 
 public interface BackupPresenter extends BasePresenter<BackupView>, BackupNextStepListener {
 
+	void onSaveInstanceState(Bundle outState);
+
+	void saveKeyData(String key);
 }
--- a/recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenter.java
+++ b/recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenter.java
@@ -1,13 +1,13 @@
 package com.kin.ecosystem.recovery.backup.presenter;
 
 import android.os.Bundle;
-import com.kin.ecosystem.recovery.backup.view.BackupNextStepListener;
+import com.kin.ecosystem.recovery.backup.view.BackupNavigator;
 import com.kin.ecosystem.recovery.backup.view.BackupView;
 import com.kin.ecosystem.recovery.base.BasePresenter;
 
-public interface BackupPresenter extends BasePresenter<BackupView>, BackupNextStepListener {
+public interface BackupPresenter extends BasePresenter<BackupView>, BackupNavigator {
 
 	void onSaveInstanceState(Bundle outState);
 
-	void saveKeyData(String key);
+	void setAccountKey(String key);
 }
--- a/recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenterImpl.java
+++ b/recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenterImpl.java
@@ -10,36 +10,40 @@ import com.kin.ecosystem.recovery.events.CallbackManager;
 
 public class BackupPresenterImpl extends BasePresenterImpl<BackupView> implements BackupPresenter {
 
-	private static final String KEY_STEP = "kinrecovery_step";
-	public static final String KEY_ACCOUNT_KEY = "kinrecovery_account_key";
+	private static final String KEY_STEP = "kinrecovery_backup_step";
+	public static final String KEY_ACCOUNT_KEY = "kinrecovery_backup_account_key";
 
 	private @Step
 	int step;
 	private final CallbackManager callbackManager;
 	private boolean isBackupSucceed = false;
-	private final Bundle savedInstanceState;
+	private String accountKey;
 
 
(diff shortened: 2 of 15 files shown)


recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenterImpl.java
```diff
@@ -2,27 +2,38 @@
...
 
+	private static final String KEY_STEP = "kinrecovery_step";
+	public static final String KEY_ACCOUNT_KEY = "kinrecovery_account_key";
+
 	private @Step
-	int step = STEP_START;
+	int step;
 	private final CallbackManager callbackManager;
 	private boolean isBackupSucceed = false;
+	private final Bundle savedInstanceState;
```
YohayBar: Whatabout having property for KEY_STEP and KEY_ACCOUNT_KEY and not savedInstanceState.
in BackupPresenterImpl if savedInstanceState != null extract both. In onSaveInstanceState save both params in outState

recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenterImpl.java
```diff
@@ -53,11 +64,14 @@ public void setStep(@Step final int step, @Nullable Bundle data) {
 					break;
 				case STEP_SAVE_AND_SHARE:
 					if (data != null) {
-						final String key = data.getString(KEY_ACCOUNT_KEY, null);
+						final String key = getAccountKey(data);
```
YohayBar: I will remove @Nullable Bundle data from setStep and instead use String key

recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenterImpl.java
YohayBar: same as you did on SaveAndSharePresenterImpl

recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenterImpl.java
```diff
@@ -68,4 +82,19 @@ public void setStep(@Step final int step, @Nullable Bundle data) {
...
+		outState.putString(KEY_ACCOUNT_KEY, getAccountKey(savedInstanceState));
+	}
+
+	@Override
+	public void saveKeyData(String key) {
+		this.savedInstanceState.putString(KEY_ACCOUNT_KEY, key);
+	}
+
+	private String getAccountKey(@NonNull final Bundle data) {
+		return data.getString(KEY_ACCOUNT_KEY);
```
YohayBar: it might return null String

recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenter.java
```diff
@@ -1,9 +1,13 @@
...
+import android.os.Bundle;
 import com.kin.ecosystem.recovery.backup.view.BackupNextStepListener;
 import com.kin.ecosystem.recovery.backup.view.BackupView;
 import com.kin.ecosystem.recovery.base.BasePresenter;
 
 public interface BackupPresenter extends BasePresenter<BackupView>, BackupNextStepListener {
 
+	void onSaveInstanceState(Bundle outState);
+
+	void saveKeyData(String key);
```
YohayBar: setKey could be better - the save is internal detail

recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenter.java
```diff
@@ -1,9 +1,13 @@
 package com.kin.ecosystem.recovery.backup.presenter;
 
+import android.os.Bundle;
 import com.kin.ecosystem.recovery.backup.view.BackupNextStepListener;
 import com.kin.ecosystem.recovery.backup.view.BackupView;
 import com.kin.ecosystem.recovery.base.BasePresenter;
 
 public interface BackupPresenter extends BasePresenter<BackupView>, BackupNextStepListener {
 
+	void onSaveInstanceState(Bundle outState);
```
YohayBar, line 10: will that affect unit test of BackupPresenter do to  Bundle

recovery/src/main/java/com/kin/ecosystem/recovery/backup/presenter/BackupPresenter.java
YohayBar: Consider to provide API for get/set  key and step explicitly within the presenter. The view will be able to set and also get the current value in the presenter. The Activity/Fragment responsibility to persist (by calling get on these property)  or restore it (by set) when needed

recovery/src/main/java/com/kin/ecosystem/recovery/backup/view/BackupActivity.java
```diff
@@ -137,10 +163,28 @@ protected void onStop() {
...
 		}
 	}
+
+	private void setCreatePasswordFragmentAttributes(CreatePasswordFragment createPasswordFragment) {
+		createPasswordFragment.setNextStepListener(backupPresenter);
+		createPasswordFragment.setKeyboardHandler(this);
+	}
+
+	private CreatePasswordFragment getSavedCreatePasswordFragment() {
+		return (CreatePasswordFragment) getSupportFragmentManager()
```
YohayBar: potential NPE

recovery/src/main/java/com/kin/ecosystem/recovery/backup/view/BackupActivity.java
```diff
@@ -48,6 +60,8 @@ public void startBackupFlow() {
 
 		if (backupInfoFragment == null) {
 			backupInfoFragment = BackupInfoFragment.newInstance(backupPresenter);
+		} else {
```
YohayBar: Our entire usage of newInstance is broken - new instance design pattern will only work for priorities that can be  stored in bundle using setArguments. just setting properties will not be store if fragment will be recreated

YohayBar (approved):
💯

PR #191, opened 2018-11-13
Nitzan/eco 758/fix webview mailto

nitzantomer:
Our webview did not support `mailto` links.
This PR fixes that.


phips28:
pls also merge it into `no_loc` branch when its ready for Swelly release. Thx!

sdk/src/main/java/com/kin/ecosystem/web/EcosystemWebViewClient.java
```diff
@@ -24,4 +31,29 @@ public void onPageFinished(WebView view, String url) {
     public void onReceivedError(WebView view, WebResourceRequest request, WebResourceError error) {
 		Logger.log(new Log().withTag(TAG).text("onReceivedError(WEB_VIEW, REQUEST, \"" + error.toString() + "\")"));
     }
+
+    @Override
+    public boolean shouldOverrideUrlLoading(WebView view, String url) {
```
YohayBar, line 36: From documentation - 
This method was deprecated in API level 24.
Use shouldOverrideUrlLoading(WebView, WebResourceRequest) instead.

YohayBar (approved):
👍 fix conflicts

sdk/src/main/java/com/kin/ecosystem/web/EcosystemWebViewClient.java
```diff
@@ -24,4 +31,29 @@ public void onPageFinished(WebView view, String url) {
...
+        if (url.startsWith("mailto:")) {
+            MailTo mt = MailTo.parse(url);
+            Intent intent = newEmailIntent(context, mt.getTo(), mt.getSubject(), mt.getBody(), mt.getCc());
+            context.startActivity(intent);
+
+            return true;
+        } else {
+            view.loadUrl(url);
```
YohayBar: ?

YohayBar: "Note: Do not call WebView.loadUrl(String) with the same URL and then return true. This unnecessarily cancels the current load and starts a new load with the same URL. The correct way to continue loading a given URL is to simply return false, without calling WebView.loadUrl(String)."

YohayBar (approved):

PR #196, opened 2018-11-16
fix crash of orderhistory don't keep activities

OrenZak:
#### Main purpose:
Crash after activity restart due to `don't keep activities` (low memory devices)
#### Technical description:
Save screen id on saveInstanceState and use it on restart activity

### DIFF
--- a/sdk/src/main/java/com/kin/ecosystem/main/view/EcosystemActivity.java
+++ b/sdk/src/main/java/com/kin/ecosystem/main/view/EcosystemActivity.java
@@ -8,6 +8,8 @@ import static com.kin.ecosystem.main.Title.ORDER_HISTORY_TITLE;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.v4.app.FragmentManager.BackStackEntry;
+import android.support.v4.app.FragmentTransaction;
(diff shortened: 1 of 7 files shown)


sdk/src/main/java/com/kin/ecosystem/main/view/EcosystemActivity.java
```diff
@@ -247,6 +267,19 @@ public void navigateBack() {
 			super.onBackPressed();
 			overridePendingTransition(0, R.anim.kinecosystem_slide_out_right);
 		} else {
+			BackStackEntry entry = getSupportFragmentManager().getBackStackEntryAt(count - 1);
+			if (entry != null && entry.getName().equals(MARKETPLACE_TO_ORDER_HISTORY)) {
+				// After pressing back from OrderHistory, should put the attrs again.
+				// This is the only fragment that should set presenter again on back.
+				MarketplaceFragment marketplaceFragment = getSavedMarketplaceFragment();
+				if (marketplaceFragment != null) {
```
YohayBar, line 275: what happens if marketplaceFragment == null?

OrenZak: no sure it's even possible, only this fragment goes to back stack.

YohayBar (approved):

PR #202, opened 2018-11-22
Eco/hotfix/api remove observer wrong method impl

OrenZak:


core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -372,8 +372,13 @@ public void updateActiveAccount(int accountIndex) throws BlockchainException {
 		local.setAccountIndex(accountIndex);
 		createKinAccountIfNeeded();
 
-		balanceRegistration.remove();
-		startBalanceListener();
+		if (balanceRegistration != null) {
+			balanceRegistration.remove();
```
YohayBar: synchronize when checking balanceObserversCount?

OrenZak: 👍

YohayBar: maybe add removeRegistration(balanceRegistration); into synchronized (balanceObserversLock) { } instead

YohayBar (approved):

PR #204, opened 2018-11-25
Crash getting null Error from server

OrenZak:
#### Main purpose:
Solve crash of getting Null Error, so we can't get the message.
Not sure where is came from client/server.
#### Technical description:
- Added an Error object if `deserializeError` failed from severs so we will know later.
- Added more error code from the server to `ErrorUtil` so we will handle all the error codes.
- Return a default error from the server, so we won't crash.
- Make the `ServiceException.getMessage()` to be the message from the server.

### DIFF
--- a/core/src/main/java/com/kin/ecosystem/core/network/ApiClient.java
+++ b/core/src/main/java/com/kin/ecosystem/core/network/ApiClient.java
@@ -576,7 +576,7 @@ public class ApiClient {
             return (Error) (json.deserialize(respBody, new TypeToken<Error>() {
             }.getType()));
         } catch (Throwable throwable) {
-            return null;
+            return new Error("Could not deserialize Error", "Could not deserialize: " + respBody, 0);
         }
     }
 
--- a/core/src/main/java/com/kin/ecosystem/core/network/model/Error.java
+++ b/core/src/main/java/com/kin/ecosystem/core/network/model/Error.java
@@ -22,11 +22,11 @@ import java.util.Objects;
  */
 public class Error {
     @SerializedName("error")
-    private String error = null;
+    private String error;
     @SerializedName("message")
-    private String message = null;
+    private String message;
     @SerializedName("code")
-    private Integer code = null;
+    private Integer code;
 
     public Error(String error, String message, Integer code) {
         this.error = error;
@@ -46,7 +46,7 @@ public class Error {
      * @return error
      **/
     public String getError() {
-        return error;
+        return error == null ? "" : error;
     }
 
     public void setError(String error) {
@@ -65,7 +65,7 @@ public class Error {
      * @return message
      **/
     public String getMessage() {
-        return message;
+        return message == null ? "" : message;
     }
 
     public void setMessage(String message) {
@@ -78,7 +78,7 @@ public class Error {
     }
 
 
-    /**
+    /**g
      * Get code
      *
      * @return code
@@ -107,6 +107,7 @@ public class Error {
 
     @Override
     public int hashCode() {
+
         return Objects.hash(error, message, code);
     }
 }
--- a/core/src/main/java/com/kin/ecosystem/core/network/model/Error.java
+++ b/core/src/main/java/com/kin/ecosystem/core/network/model/Error.java
@@ -21,6 +21,9 @@ import java.util.Objects;
  * an error
  */
 public class Error {
+
+    private static final String EMPTY_STR = "";
+
     @SerializedName("error")
     private String error;
     @SerializedName("message")
@@ -46,7 +49,7 @@ public class Error {
      * @return error
      **/
     public String getError() {
-        return error == null ? "" : error;
+        return error == null ? EMPTY_STR : error;
     }
 
     public void setError(String error) {
@@ -65,7 +68,7 @@ public class Error {
      * @return message
      **/
     public String getMessage() {
-        return message == null ? "" : message;
+        return message == null ? EMPTY_STR : message;
     }
 
     public void setMessage(String message) {
@@ -78,7 +81,7 @@ public class Error {
     }
 
 
-    /**g
+    /**
      * Get code
      *
      * @return code
--- a/core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
+++ b/core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
@@ -17,7 +17,6 @@ import kin.core.exception.TransactionFailedException;
 public class ErrorUtil {
 
 	// Error messages
-	private static final String USER_NOT_FOUND_ON_ECOSYSTEM_SERVER = "User not found on the Ecosystem server";
 	private static final String THE_ECOSYSTEM_SERVER_RETURNED_AN_ERROR = "The Ecosystem server returned an error. See underlying Error for details";
 	private static final String ECOSYSTEM_SDK_ENCOUNTERED_AN_UNEXPECTED_ERROR = "Ecosystem SDK encountered an unexpected error";
 	private static final String BLOCKCHAIN_ENCOUNTERED_AN_UNEXPECTED_ERROR = "Blockchain encountered an unexpected error";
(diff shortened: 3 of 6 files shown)


sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/SpendDialogPresenter.java
```diff
@@ -91,10 +91,10 @@ public void onResponse(OpenOrder response) {
 			public void onFailure(KinEcosystemException exception) {
 				showToast(SOMETHING_WENT_WRONG);
 				try {
-					String errorMsg = ((ApiException) exception.getCause()).getResponseBody().getMessage();
+					String errorMsg = exception.getMessage();
 					eventLogger.send(SpendOrderCreationFailed
 						.create(errorMsg, offer.getId(), false, SpendOrderCreationFailed.Origin.MARKETPLACE));
-				} catch (ClassCastException e) {
+				} catch (Throwable e) {
```
YohayBar: no need to try catch both lines are identical and use exception.getMessage() as the message - this will also really help with catching crash as the exception.getMessage() is used again

core/src/main/java/com/kin/ecosystem/core/network/model/Error.java
```diff
@@ -65,7 +65,7 @@ public Error message(String message) {
      * @return message
      **/
     public String getMessage() {
-        return message;
+        return message == null ? "" : message;
```
YohayBar: change ""  to public CONST something like " Missing error message" in case we get Error event without message we will know the origin of the missing message

core/src/main/java/com/kin/ecosystem/core/network/ApiClient.java
```diff
@@ -576,7 +576,7 @@ private Error deserializeError(String respBody) {
             return (Error) (json.deserialize(respBody, new TypeToken<Error>() {
             }.getType()));
         } catch (Throwable throwable) {
-            return null;
+            return new Error("Could not deserialize Error", "Could not deserialize: " + respBody, 0);
```
YohayBar, line 579: Consider adding a second parameter i.e deserializeError(String respBody, String origin)  the origin will be used to track where deserializeError called from

YohayBar: could be passed as the error code

core/src/main/java/com/kin/ecosystem/core/network/model/Error.java
```diff
@@ -78,7 +78,7 @@ public Error code(Integer code) {
     }
 
 
-    /**
+    /**g
```
YohayBar: remove g

core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
```diff
@@ -39,31 +38,32 @@
...
+				case ERROR_CODE_TRANSACTION_FAILED_ERROR:
+					Error error = apiException.getResponseBody();
+					if (error != null) {
+						String msg = error.getMessage();
+						exception = new ServiceException(ServiceException.SERVICE_ERROR,
+							(msg != null && !msg.isEmpty()) ? msg : THE_ECOSYSTEM_SERVER_RETURNED_AN_ERROR,
+							apiException);
+						break;
+					}
+					exception = createUnknownServiceException(apiException);
```
yosriz, line 66: seems like a typo, you overwriting the extracted exception

sdk/src/main/java/com/kin/ecosystem/poll/presenter/PollWebViewPresenter.java
```diff
@@ -92,13 +92,7 @@ public void onResponse(OpenOrder response) {
...
 			public void onFailure(KinEcosystemException exception) {
 				showToast(SOMETHING_WENT_WRONG);
-				String errorMsg;
-				try {
-					errorMsg = ((ApiException) exception.getCause()).getResponseBody().getMessage();
-				} catch (Throwable error) {
-					errorMsg = error.getMessage();
-				}
-
+				String errorMsg = exception.getMessage();
```
yosriz, line 95: I think the best way for handling unexpected errors in log is to print the entire stacktrace to get as much data as needed using android `Log.getStackTraceString()`, it will print also the cause.
You can see an example here https://github.com/kinecosystem/atn-android/blob/0e4c0269783a82c2bfa1773159cd239ea530ef06/atnlib/src/main/java/com/kik/atn/AndroidLogger.java#L57 (I did a fix there aa Android don't print `UnknownHostException` and we did want to know about it)

core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
```diff
@@ -39,31 +38,32 @@
 	public static final int ERROR_CODE_REQUEST_TIMEOUT = 408;
 	public static final int ERROR_CODE_CONFLICT = 409;
 	public static final int ERROR_CODE_INTERNAL_SERVER_ERROR = 500;
+	public static final int ERROR_CODE_TRANSACTION_FAILED_ERROR = 700;
 
-	public static final int ERROR_CODE_USER_NOT_FOUND = 4046;
```
YohayBar, line 43: don't we have 4046 error code?

core/src/main/java/com/kin/ecosystem/core/network/model/Error.java
YohayBar: you can also just provide a default value for the property

OrenZak: i don't want to use default value because it can be set to null.
i'll use const

core/src/main/java/com/kin/ecosystem/core/network/ApiClient.java
OrenZak: not sure why do we need it.
it can be called only from the client.

core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
OrenZak: I'm not, there is a break before.

yosriz: :eyes: Ohh, that's confusing

sdk/src/main/java/com/kin/ecosystem/poll/presenter/PollWebViewPresenter.java
OrenZak: I'm getting the error msg its enough for us at this moment, this is not the case we want to see the stacktrace. later on, we will have an exception handler that will post the stacktrace to our endpoint, it's an open task :)

core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
OrenZak: we did, but it's part of `ERROR_CODE_NOT_FOUND` so we handled it specifically but I'm taking the message from the `apiException` so it will be a more clear message and not default one.

sdk/src/main/java/com/kin/ecosystem/poll/presenter/PollWebViewPresenter.java
yosriz: :+1:

core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
```diff
@@ -74,16 +74,16 @@ public static KinEcosystemException fromApiException(ApiException apiException)
...
 			}
 		}
 		return exception;
 	}
 
-	private static KinEcosystemException createUnknownException(@Nullable Throwable throwable) {
-		return new KinEcosystemException(KinEcosystemException.UNKNOWN, ECOSYSTEM_SDK_ENCOUNTERED_AN_UNEXPECTED_ERROR,
-			throwable);
+	private static KinEcosystemException createUnknownServiceException(@Nullable Throwable throwable) {
+		final String msg  = (throwable != null && throwable.getMessage() != null) ? throwable.getMessage() : ECOSYSTEM_SDK_ENCOUNTERED_AN_UNEXPECTED_ERROR;
```
YohayBar: if throwable.getMessage() == null we might still get valuble inforamtion for example from getCause() - meaing we have undeling cause with more information.

core/src/main/java/com/kin/ecosystem/core/util/ErrorUtil.java
```diff
@@ -39,31 +38,32 @@
...
 				case ERROR_CODE_INTERNAL_SERVER_ERROR:
-					exception = new ServiceException(ServiceException.SERVICE_ERROR,
-						THE_ECOSYSTEM_SERVER_RETURNED_AN_ERROR,
-						apiException);
+				case ERROR_CODE_TRANSACTION_FAILED_ERROR:
+					Error error = apiException.getResponseBody();
+					if (error != null) {
+						String msg = error.getMessage();
+						exception = new ServiceException(ServiceException.SERVICE_ERROR,
+							(msg != null && !msg.isEmpty()) ? msg : THE_ECOSYSTEM_SERVER_RETURNED_AN_ERROR,
```
YohayBar, line 62: if a message is null try we can try to get apiException.getMessage()

OrenZak: it is like that see  `createUnknownServiceException` method

YohayBar (approved):

PR #205, opened 2018-11-25
Recover from process restart

OrenZak:
#### Main purpose:
Crash after Android OS has restarted the application process.
#### Technical description:
- Create InitProvider, so we can initiate all our repos on app start.
- Separate start into: `initialize` and `login` method, where the `initialize` are being called by our InitProvider and `login` by the developer.
- There are no network calls in `initialize`.
- Added login callback so the developer will have the ability to know when he can show the kin icon/ interact with the user account.
#### Dilemmas you faced with? 
How to init all our repos on app start without the developer to do something.
#### Tests:
Update the tests to the current implementation.
#### Documentation:
Updated according to the current implementation.


sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -74,27 +84,106 @@ private static Kin getInstance() {
 		return instance;
 	}
 
-	public static void start(@NonNull Context appContext, @NonNull WhitelistData whitelistData,
-		@NonNull KinEnvironment environment)
-		throws ClientException, BlockchainException {
+	synchronized static void initialize(Context appContext) throws ClientException {
```
YohayBar, line 87: does it need to be synchronized?

app/src/main/AndroidManifest.xml
```diff
@@ -14,6 +14,9 @@
         android:theme="@style/AppTheme"
         android:name=".App">
 
+        <meta-data android:name="com.kin.ecosystem.sdk.EnvironmentName" android:value="beta"/>
```
YohayBar: It will be safer is name will be preset as a string resource 
```
<meta-data
    android:name="com.facebook.sdk.ApplicationId"
    android:value="@string/kinecosystem_enviorment_beta" />
```
You can define kinecosystem_enviorment_beta in strings.xml or from gradle

app/src/main/java/com/ecosystem/kin/app/App.java
```diff
@@ -2,47 +2,70 @@
...
+                    public void onResponse(Void response) {
+                        Log.d(TAG, "JWT onResponse: login");
+                    }
+
+                    @Override
+                    public void onFailure(KinEcosystemException exception) {
+                        Log.e(TAG, "JWT onFailure: " + exception.getMessage());
+                    }
+                });
+            } catch (BlockchainException e) {
```
YohayBar: why do we need both try-catch for BlockchainException if we have onFailure callback?

app/src/main/java/com/ecosystem/kin/app/App.java
```diff
@@ -2,47 +2,70 @@
...
+                    public void onResponse(Void response) {
+                        Log.d(TAG, "WhiteList onResponse: login");
+                    }
+
+                    @Override
+                    public void onFailure(KinEcosystemException exception) {
+                        Log.e(TAG, "WhiteList onFailure: " + exception.getMessage());
+                    }
+                });
+            } catch (BlockchainException e) {
```
YohayBar: same

core/src/main/java/com/kin/ecosystem/core/accountmanager/AccountManagerImpl.java
```diff
@@ -152,6 +159,7 @@ public void onResponse(Void response) {
 
 						@Override
 						public void onFailure(KinEcosystemException error) {
+							instance.error = error;
```
YohayBar, line 162: also in switchAccount?

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthLocalData.java
```diff
@@ -60,6 +60,23 @@ public void setSignInData(@NonNull final SignInData signInData) {
...
+	public SignInData getSignInData() {
+		final String deviceID = signInSharedPreferences.getString(DEVICE_ID_KEY, null);
+		final String publicAddress = signInSharedPreferences.getString(PUBLIC_ADDRESS_KEY, null);
+		final String signInType = signInSharedPreferences.getString(TYPE_KEY, null);
+		final String jwt = signInSharedPreferences.getString(JWT_KEY, null);
+		final String userID = signInSharedPreferences.getString(USER_ID_KEY, null);
+		final String appID = signInSharedPreferences.getString(APP_ID_KEY, null);
+		if (deviceID != null && publicAddress != null && signInType != null && jwt != null && userID != null
+			&& appID != null) {
+			return new SignInData().deviceId(deviceID).walletAddress(publicAddress)
```
YohayBar, line 73: you can serialize and descrilze SignInData to/from json string and get/set signInSharedPreferences once

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
```diff
@@ -33,6 +33,8 @@ private AuthRepository(@NonNull AuthDataSource.Local local,
 		@NonNull AuthDataSource.Remote remote) {
 		this.localData = local;
 		this.remoteData = remote;
+		this.cachedSignInData = local.getSignInData();
```
YohayBar, line 36: should happen but just to be on the safe side also check cachedSignInData != null in updateWalletAddress()

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -98,20 +98,31 @@ public static BlockchainSourceImpl getInstance() {
...
 	}
 
-	private void createKinAccountIfNeeded() throws BlockchainException {
-		int accountIndex = local.getAccountIndex();
+	private void initAccount() {
 		if (kinClient.hasAccount()) {
-			account = kinClient.getAccount(accountIndex);
-		} else {
+			if (account == null) {
+				account = kinClient.getAccount(activeAccountIndex);
```
YohayBar: get account could be  a bit heavy operation when the app just starts

YohayBar: do we have to get the account in init phase?

sdk/src/main/java/com/kin/ecosystem/Kin.java
OrenZak: yes to be on the safe side.

app/src/main/AndroidManifest.xml
OrenZak: 👍

app/src/main/java/com/ecosystem/kin/app/App.java
OrenZak: nice catch :)

core/src/main/java/com/kin/ecosystem/core/accountmanager/AccountManagerImpl.java
OrenZak: @YohayBar not sure i understand why do you mean ?

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
OrenZak: there is no need, you already logged in and just updating wallet address.

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthLocalData.java
OrenZak: ?

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: nop, i'll move it into createAccountIfNeeded

app/src/main/AndroidManifest.xml
```diff
@@ -12,7 +12,11 @@
         android:label="@string/app_name"
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
-        android:name=".App">
+        android:name=".App"
+        tools:ignore="GoogleAppIndexingWarning">
```
YohayBar, line 16: ?

core/src/main/java/com/kin/ecosystem/core/accountmanager/AccountManagerImpl.java
YohayBar: Consider set error in onFailure within the switch account

sdk/src/main/java/com/kin/ecosystem/Kin.java
YohayBar: it is not safe - we can create ANR - we can only call it from this package and we are only calling it from KinEcosystemInitProvider. using synchronized have some performance cost and we shouldn't use it when called from Provider. There is another issue that I just thought about now (sorry it just now)- what if the developer wants to test the SDK on a small set of its user? if we place init in KinEcosystemInitProvider this code will run on each instance whether or not it is part of the AB.

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -109,89 +198,68 @@ private static SignInData getJwtSignInData(@NonNull final String jwt) {
...
+								public void run() {
+									loginCallback.onFailure(AccountManagerImpl.getInstance().getError());
+								}
+							});
+							break;
+					}
+				}
+			});
+		} else {
+			isAccountLoggedIn = true;
```
yosriz, line 260: what happens in case that account is already created? does the `addAccountStateObserver` will trigger  `loginCallback.onResponse`? it seems that it will never be called in this case? (cause you will not call `AccountManagerImpl.getInstance().start();`)

sdk/src/main/java/com/kin/ecosystem/Kin.java
yosriz: I also agree, and this is a use case I didn't even think about, but generally, this kind of SDK, which has nothing special to manage across the entire application lifecycle, is not in a position to force itself whenever you take it. I think that developer should choose whenever he wants to init it. (I'm also not a fan of generally expose it as static from similar reasons)

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -74,27 +84,105 @@ private static Kin getInstance() {
...
 		Logger.enableLogs(enableLogs);
 	}
 
+	public static void login(@NonNull WhitelistData whitelistData, KinCallback<Void> loginCallback) {
+		SignInData signInData = getWhiteListSignInData(whitelistData);
+		init(signInData, loginCallback);
+
+	}
+
+	public static void login(@NonNull String jwt, KinCallback<Void> loginCallback) {
```
yosriz, line 181: :clap::clap: Login callback!!

PR #208, opened 2018-11-27
DP-149 - fix a problem with importing some QRs

yosriz:
#### Main purpose:
Fix an issue where some QRs cannot be imported and the user gets import error in restore flow.
#### Technical description:
* add TRY_HARDER, PURE_BARCODE flags to make sure we put max efforts decoding a QR...
* use QRCodeReader as it supposed to be faster
* add a test with a QR that cannot be decoded before
following this issue in zxing lib https://github.com/zxing/zxing/issues/995 
(thanks @steelbytes!)
#### Dilemmas you faced with? 
N/A
#### Tests:
add a test with a QR that cannot be decoded before
#### Documentation:
N/A


recovery/src/androidTest/java/com/kin/ecosystem/recovery/qr/QRBarcodeGeneratorImplTest.java
```diff
@@ -73,10 +79,16 @@ public void generate_success() throws Exception {
 
 	@Test
 	public void decodeQR_success() throws Exception {
-		Bitmap bitmap = TestUtils.loadBitmapFromResource(this.getClass(), "qr_test.png");
+		// test a QR that was being able to be decoded only with TRY_HARDER flag on in zxing library
+		Bitmap bitmap = TestUtils.loadBitmapFromResource(this.getClass(), "backup_qr_try_harder.png");
```
YohayBar, line 83: backup_qr_try_harder.png is the QR that didn't succeed before?

YohayBar (approved):
look good to me - need to also QA test it

recovery/src/androidTest/java/com/kin/ecosystem/recovery/qr/QRBarcodeGeneratorImplTest.java
yosriz: yep

OrenZak:
@yosriz please sync with dev

PR #209, opened 2018-11-29
make sure list not null before sync

OrenZak:
#### Main purpose:
old list was null, on an edge case of receiving Offers from the server and going to order history so the presenter being detached.
#### Technical description:
- make sure old list is not null by creating new ArrayList if it's null
- removing the order observer before assigning the `earnList` and `spendList` to null `onDetach`.

### DIFF
--- a/sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/MarketplacePresenter.java
+++ b/sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/MarketplacePresenter.java
@@ -203,9 +203,9 @@ public class MarketplacePresenter extends BasePresenter<IMarketplaceView> implem
 	}
 
 	private void release() {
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/MarketplacePresenter.java
```diff
@@ -244,6 +244,9 @@ private void syncOffers(OfferList offerList) {
 	}
 
 	private void syncList(List<Offer> newList, List<Offer> oldList, OfferType offerType) {
+		// Make sure olsList is not null
+		if (oldList == null) { oldList = new ArrayList<>(); }
```
YohayBar: check before calling syncList that oldList!= null

YohayBar (approved):

YohayBar (approved):

PR #212, opened 2018-12-04
Eco 661/marketplace leak

OrenZak:
#### Main purpose:
Prevent as much as we can memory leaks
#### Technical description:
- Release all of the resources on destroy and when there is no use for them anymore.
- Added LeakCanary to be aware of leaks in the future :)
#### Tests:
- Tested on a variety of phones
- Tested with the android profiler 
- Tested with `adb dumpsys meminfo`

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -80,6 +80,9 @@ dependencies {
         transitive = true;
     }
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
+    debugImplementation 'com.squareup.leakcanary:leakcanary-android:1.6.2'
+    debugImplementation 'com.squareup.leakcanary:leakcanary-support-fragment:1.6.2'
 }
 
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -81,6 +81,7 @@ dependencies {
     }
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
+    releaseImplementation 'com.squareup.leakcanary:leakcanary-android-no-op:1.6.2'
     debugImplementation 'com.squareup.leakcanary:leakcanary-android:1.6.2'
     debugImplementation 'com.squareup.leakcanary:leakcanary-support-fragment:1.6.2'
 }
--- a/recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
+++ b/recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
@@ -5,7 +5,6 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
-import android.support.annotation.IntDef;
 import android.support.annotation.NonNull;
 import android.support.annotation.StringDef;
 import android.support.v4.content.LocalBroadcastManager;
@@ -48,16 +47,17 @@ public class BroadcastManagerImpl implements BroadcastManager {
 
 	@Override
 	public void unregister() {
-		if (receiver != null) {
+		if (receiver != null && activity != null) {
 			LocalBroadcastManager.getInstance(activity).unregisterReceiver(receiver);
 		}
-		activity = null;
 	}
 
 	@Override
 	public void sendEvent(Intent data, @ActionName final String actionName) {
-		data.setAction(actionName);
-		LocalBroadcastManager.getInstance(activity).sendBroadcast(data);
+		if(activity != null) {
+			data.setAction(actionName);
+			LocalBroadcastManager.getInstance(activity).sendBroadcast(data);
+		}
 	}
 
 	@Override
@@ -65,5 +65,12 @@ public class BroadcastManagerImpl implements BroadcastManager {
 		if (activity != null) {
 			activity.setResult(resultCode, data);
 		}
+	}
+
+	@Override
+	public void release() {
+		// Safe unregister
+		unregister();
+		activity = null;
 	}
 }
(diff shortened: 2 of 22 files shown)


recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
```diff
@@ -51,6 +51,7 @@ public void unregister() {
 		if (receiver != null) {
 			LocalBroadcastManager.getInstance(activity).unregisterReceiver(receiver);
 		}
+		activity = null;
```
YohayBar: how can you be sure that BroadcastManagerImpl is not reused - meaning same BroadcastManagerImpl is registered to,  unregister and register again

YohayBar: activity is also used in more places and not protected against NPE

sdk/src/main/java/com/kin/ecosystem/history/view/OrderHistoryFragment.java
```diff
@@ -37,9 +37,13 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
 	@Override
-	public void onDestroyView() {
-		super.onDestroyView();
-		orderHistoryPresenter.onDetach();
+	public void onDestroy() {
+		if (orderHistoryPresenter != null) {
+			orderHistoryPresenter.onDetach();
+			orderHistoryPresenter = null;
+		}
+		orderHistoryRecyclerAdapter = null;
```
YohayBar, line 45: protect orderHistoryRecyclerAdapter in other places - related methods could be called from async operations even after onDestroy

sdk/src/main/java/com/kin/ecosystem/main/presenter/EcosystemPresenter.java
```diff
@@ -85,6 +85,8 @@ public void onSaveInstanceState(Bundle outState) {
 	public void onDetach() {
 		super.onDetach();
 		removeBalanceObserver();
+		navigator = null;
```
YohayBar, line 92: navigator is provided in the constructor and not in onAttached. can EcosystemPresenter be reused after onDetach.

protect against NPE in case related methods are called from async operations

app/build.gradle
```diff
@@ -80,6 +80,10 @@ dependencies {
         transitive = true;
     }
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
+    releaseImplementation 'com.squareup.leakcanary:leakcanary-android-no-op:1.6.2'
```
yosriz, line 84: consider using leak canary in release in order to include it in QA tests

sdk/src/main/java/com/kin/ecosystem/history/view/OrderHistoryFragment.java
```diff
@@ -37,9 +37,13 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
 	}
 
 	@Override
-	public void onDestroyView() {
-		super.onDestroyView();
-		orderHistoryPresenter.onDetach();
+	public void onDestroy() {
+		if (orderHistoryPresenter != null) {
```
yosriz, line 41: don't see a particular reason to null out presenter, the leak direction is the opposite when presenter holds the activity, isn't it?

sdk/src/main/java/com/kin/ecosystem/history/view/OrderHistoryFragment.java
yosriz: and why you need to null out adapter?

yosriz: I think the approach should be to null out the view in base presenter, and check in each presenter for a null view before each operation.

yosriz:
@OrenZak I don't understand what was the leak source? did you find it? was it the Observer and nulling it out solved the issue?

recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
OrenZak: 1. The unregister is being call from `BackupManager.release()` meaning the you are not going to use the same `BackupManager` again, so it should be initialized again for use.
2. where ?

sdk/src/main/java/com/kin/ecosystem/history/view/OrderHistoryFragment.java
OrenZak: 1. the adapter contains the recyclerview the contains the activity as a context, so in some scenario is can be leaked.
2. it is like that but it doesn't mean that something else does not save the activity instance.

sdk/src/main/java/com/kin/ecosystem/main/presenter/EcosystemPresenter.java
OrenZak: `ecosystemPresenter.onDetach()` is being called from view `onDestroy` so it can't be reused.

app/build.gradle
OrenZak: QA tests on debug first :)

recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
YohayBar: Could happen I how could you be sure it will not happen in the future

sdk/src/main/java/com/kin/ecosystem/history/view/OrderHistoryFragment.java
OrenZak: first of all release memory it's a good approach :), but you are right the leak direction is the opposite.

yosriz (approved):
:white_check_mark: no more leaks!

PR #218, opened 2018-12-11
crash on native offer clicked after few times

OrenZak:
#### Main purpose:
Crash after few clicks on native offers
#### Technical description:
- add onStart, onStop for clear observers
- onAttach and onDetach are being called once now, so it can't be null now.

### DIFF
--- a/sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/MarketplacePresenter.java
+++ b/sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/MarketplacePresenter.java
@@ -77,10 +77,30 @@ public class MarketplacePresenter extends BasePresenter<IMarketplaceView> implem
 	@Override
 	public void onAttach(IMarketplaceView view) {
 		super.onAttach(view);
+		eventLogger.send(MarketplacePageViewed.create());
+	}
+
+	@Override
+	public void onDetach() {
+		super.onDetach();
+		navigator = null;
+	}
+
+	@Override
+	public void onStart() {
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/MarketplacePresenter.java
```diff
@@ -77,10 +77,30 @@ public MarketplacePresenter(@NonNull final IMarketplaceView view, @NonNull final
 	@Override
 	public void onAttach(IMarketplaceView view) {
 		super.onAttach(view);
+		eventLogger.send(MarketplacePageViewed.create());
+	}
+
+	@Override
+	public void onDetach() {
+		super.onDetach();
+		navigator = null;
```
YohayBar, line 86: you shouldn't set navigator = null -> the scope of MarketplacePresenter.java is handled by EcosystemActivity navigator and shouldn't be set to null if MarketplacePresenter is still usable from EcosystemActivity.
BAse on symmetry principal it should be set to null in the same scope for example set in onStart and set to null in onStop in same place

OrenZak: it's not true, because the marketplacePresenter can hold an instance for the activity.
So when the activity is destroyed, (going back from Marketplace), the presenter still hold the activity instance and leak it. I can see it on LeakCanary.

YohayBar (approved):

PR #226, opened 2018-12-16
DP-108 - activity leaks

yosriz:
#### Main purpose:
Fix leaks (Activities and everything else) sourced at `recovery` module.
#### Technical description:
* `BroadcastManagerImpl` - each register call adds a new
receiver to the Android's LocalBroadcastManager which is static module, while
in unregister we only clear a single receiver, thus leaking the BroadcastManagerImpl
and eventually the activity that its holds in a reference.
* `CreatePassword/RestoreEnterPasswordFragment` - TextWatcherAdapter used Timer that did not cancelled and thus caused a leak
(TimerTask uses Thread internally and this thread kept reference to the callback,
to the holding fragment and so on), Timer is not needed at all, and handler can
be used.
As a "release" method is needed on destory to remove handler callbacks,
that requires keeping TextWatcherAdapter refernce, I removed TextWatcherAdapter
and make everything local.
#### Dilemmas you faced with? 
I think there's no need for nulling out references, that may obscure internal objects that are leaked. it's preferable to let the activity leak in this case, and we have more chance catch it by leak canary, instead of falsly having no activity leaked but internl objects.
#### Tests:
Manually, used LeakCanary and `dumpsys meminfo` as the real source to check after everything closed that no acitvity or view was leaked.
#### Documentation:
N/A


recovery/src/main/java/com/kin/ecosystem/recovery/backup/view/AbstractTextWatcher.java
```diff
@@ -0,0 +1,23 @@
+package com.kin.ecosystem.recovery.backup.view;
+
+import android.text.Editable;
+import android.text.TextWatcher;
+
+
+public class AbstractTextWatcher implements TextWatcher {
```
OrenZak: Delete and use the Adapter with Handler inside.

yosriz: fixed :ok_hand:

recovery/src/main/java/com/kin/ecosystem/recovery/backup/view/TextWatcherAdapter.java
```diff
@@ -32,22 +29,16 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
...
-						@Override
-						public void run() {
-							listener.afterTextChanged(editable);
-						}
-					});
-				}
-			},
-			DELAY
-		);
+		mainThreadHandler.postDelayed(new Runnable() {
```
OrenZak, line 33: you need to cancel the Runnable before each post.

recovery/src/main/java/com/kin/ecosystem/recovery/restore/view/RestoreEnterPasswordFragment.java
```diff
@@ -163,4 +165,11 @@ public void invalidQrError() {
 		contentText.setTextColor(ContextCompat.getColor(getContext(), R.color.kinrecovery_red));
 		password.setFrameBackgroundColor(R.color.kinrecovery_red);
 	}
+
+	@Override
+	public void onDestroy() {
+		super.onDestroy();
+
+		textWatcherAdapter.release();
```
OrenZak, line 173: should be called before super.

recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
```diff
@@ -31,46 +29,38 @@
...
-		LocalBroadcastManager.getInstance(activity).registerReceiver(receiver, new IntentFilter(actionName));
+		receiversList.add(broadcastReceiver);
+		LocalBroadcastManager.getInstance(activity).registerReceiver(broadcastReceiver, new IntentFilter(actionName));
 	}
 
 	@Override
-	public void unregister() {
-		if (receiver != null && activity != null) {
-			LocalBroadcastManager.getInstance(activity).unregisterReceiver(receiver);
+	public void unregisterAll() {
```
OrenZak, line 48: should we add option to delete a listener, and not all of them ?

recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
```diff
@@ -31,46 +29,38 @@
 
 	public BroadcastManagerImpl(@NonNull Activity activity) {
 		this.activity = activity;
+		receiversList = new ArrayList<>();
```
OrenZak: this. receiversList ?

recovery/src/main/java/com/kin/ecosystem/recovery/backup/view/CreatePasswordFragment.java
```diff
@@ -205,4 +212,12 @@ public void onClick(DialogInterface dialog, int which) {
 			.create()
 			.show();
 	}
+
+	@Override
+	public void onDestroy() {
+		super.onDestroy();
+
+		confirmPassTextWatcherAdapter.release();
```
OrenZak, line 220: before super ?

OrenZak:
sync with dev please

recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
yosriz: as it's internal mechanism let's add it by demand in the future

recovery/src/main/java/com/kin/ecosystem/recovery/backup/view/TextWatcherAdapter.java
yosriz: :+1:, sure

recovery/src/main/java/com/kin/ecosystem/recovery/restore/view/RestoreEnterPasswordFragment.java
yosriz: why? don't think there is some particular guide regarding that.
From looking at base class (support v4 fragment) super.onDestory merely releasing loader if any (which we don't have, so it does nothing), also looking at `AppCompatActivity` code, there's nothing I can say that must happen at the end. (removing internal handlers callbacks, flagging that activity is destoryed etc.0

recovery/src/main/java/com/kin/ecosystem/recovery/events/BroadcastManagerImpl.java
OrenZak: 👍

recovery/src/main/java/com/kin/ecosystem/recovery/restore/view/RestoreEnterPasswordFragment.java
OrenZak: Well, i've noticed that could happened that you resources won't be GC'ed.
But if you sure 👍

OrenZak:
@yosriz did you ran a test with `don't keep activities` and do a process restart test ?

OrenZak (approved):

PR #230, opened 2018-12-19
Eco/load account after process restart

OrenZak:
#### Main purpose:
The account is not loaded on init so we should call createAccount that will load that last account after process restart.


sdk/src/main/java/com/kin/ecosystem/base/KinEcosystemInitiator.java
```diff
@@ -16,7 +18,9 @@ private KinEcosystemInitiator() {
 	protected static void init(Context context) {
 		try {
 			Kin.initialize(context);
-		} catch (ClientException e) {
+			// If we had process restart we should load the account.
+			BlockchainSourceImpl.getInstance().createAccount();
```
yosriz, line 22: maybe worth changing the name to reflect it's either creating or loading the account?

yosriz (approved):

PR #232, opened 2018-12-20
update kin core to 0.1.5

OrenZak:
#### Main purpose:
Update kin core to 0.1.5


YohayBar (approved):

OrenZak:
There is a problem build kin-core with proguard, i'll wait for release from @yosriz

PR #233, opened 2018-12-20
update kin core - lazy init libsodium, multitouch on marketplace false

OrenZak:
#### Main purpose:
- Update kin core to lazy init libsodium
- catch libsodium load lib exception and show error
- disable multitouch on ecosystem main activity

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 ext {
 
     //Version
-    kinCoreVersion = '3907369a9c'
+    kinCoreVersion = '95645e1c17'
     constraintLayoutVersion = '1.1.2'
     supportVersion = '26.1.0'
     picassoVersion = '2.71828'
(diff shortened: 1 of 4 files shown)


sdk/src/main/res/layout/kinecosystem_activity_main.xml
```diff
@@ -5,6 +5,7 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:orientation="vertical"
+    android:splitMotionEvents="false"
```
YohayBar, line 8: this is the default setting?

dependencies.gradle
```diff
@@ -1,7 +1,7 @@
 ext {
 
     //Version
-    kinCoreVersion = '3907369a9c'
+    kinCoreVersion = '95645e1c17'
```
YohayBar, line 4: 95645e1c17 parent is not 3907369a9c?

OrenZak: no

sdk/src/main/res/layout/kinecosystem_activity_main.xml
OrenZak: what do you mean by default ?

YohayBar (approved):

PR #241, opened 2018-12-26
remove api progress listener

OrenZak:
#### Main purpose:
the api progress listener caused memory leaks
#### Technical description:
- remove api progress listener


yosriz (approved):
:+1:
I deeply encourage you to remove previous changes of nulling out presenters and many other objects.
this will obscure future leaks that may be introduced, and actually did obscure now the internal leaks of presenters and other objects (but not the activities)

PR #242, opened 2018-12-26
Lior/add test env

liorama:
#### Main purpose:
Provide an ability for devs and QA to test versions without interfering with partner testing.
#### Technical description:
Added a new testing environment config for an environment named (surprisingly): Test
#### Dilemmas you faced with? 
There weren't any of the usual Wednesday morning buns/bagels. I had to eat oatmeal....I guess that's not really a dilemma...
#### Tests:
I just remembered a dilemma that did come up but I don't want to move my hand and click the UP arrow so I'll write it here. Anyway, should we say anything to Ronan/Orna about telling us when no buns/bagels will be provided?
#### Documentation:
Yes, I think documentation is a good thing.


OrenZak (approved):
Great description !

PR #243, opened 2018-12-26
Add new transfer module

shayb:
The module is responsible for transferring Kin between apps.
At the moment we have implemented only the side that receives a request
to provide the current account public address and returns it back.

The request is received via an intent to a service extending
AccountInfoService. The intent contains information on how to reply back
to the source application that made the request.

#### Main purpose:
add ability to transfer kin from app to app
#### Technical description:
usage of services to wake up far app to get info and return to the caller service
#### Dilemmas you faced with? 
(Not a must)
#### Tests:
(Main tests you added/changed)
#### Documentation:
(Source/readme.md)


app/build.gradle
```diff
@@ -68,6 +68,8 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':sdk')
+    implementation project(':transfer')
```
OrenZak, line 71: Not sure we want to add it right now to the sample app, @YohayBar WDYT ?
Should be tested with QA also, please provide a version to Avishy so he can test.

app/src/main/AndroidManifest.xml
```diff
@@ -39,6 +39,16 @@
             </intent-filter>
         </activity>
 
+        <service
+            android:name=".AccountInfoServiceImpl"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="kin.transfer.info"/>
```
OrenZak: maybe add `.ecosystem` after `kin`

app/src/main/java/com/ecosystem/kin/app/AccountInfoServiceImpl.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.ecosystem.kin.app.model.SignInRepo;
+import com.kin.ecosystem.Kin;
+import com.kin.ecosystem.common.KinCallback;
+import com.kin.ecosystem.common.exception.KinEcosystemException;
+import com.kin.ecosystem.common.model.WhitelistData;
+import com.kin.ecosystem.transfer.AccountInfoService;
+/*
+*  Example usage of extending AccountInfoService
+*  Name of class should not be changed
```
OrenZak, line 11: Why ? is it explained somewhere ?

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.List;
+
+
+public abstract class AccountInfoService extends IntentService {
+    public interface AccountInfoListener {
+        void onAccountInfoReady(String data);
```
OrenZak, line 27: it's not clear that you should receive publicAddress.

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+
+        void onAccountInfoFailed(String reason);
```
OrenZak, line 29: why do we need it ?

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private static String FILE_PROVIDER_DIR_NAME = "kin_account_info";
```
OrenZak, line 32: should be with prefix of `kintransfer`, so we can the developer not to delete those file in any case.

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    protected void onHandleIntent(final @Nullable Intent intent) {
+        startForeground();
+
+        if (intent != null && intent.getAction().equals(ACTION_KIN_TRANSFER_INFO) && intent.hasExtra(EXTRA_ACTION_KEY)
+                && intent.hasExtra(EXTRA_PACKAGE_KEY) && intent.hasExtra(EXTRA_SERVICE_FULL_PATH_KEY)) {
+            getAccountInfo(new AccountInfoListener() {
+                @Override
+                public void onAccountInfoReady(String data) {
```
OrenZak, line 61: we need to check it's a valid address (length etc.)?

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+                    return false;
+                }
+            }
+            FileWriter writer = new FileWriter(file);
+            writer.append(data);
+            writer.flush();
+            writer.close();
+        } catch (IOException e) {
+            e.printStackTrace();
+            Log.e("AccountInfoService", "AccountInfoService cant create and write to file " + e.getMessage());
```
OrenZak, line 124: we can remove `e.printStackTrace();`  and Log ?
The developer can recover from that ? get any notification ?

transfer/src/test/java/com/kin/ecosystem/transfer/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
```
OrenZak, line 12: delete file, add test ?

transfer/src/main/res/xml/file_provider_paths.xml
```diff
@@ -0,0 +1,5 @@
+<paths>
+    <files-path
+        path="kin_account_info/"
```
OrenZak, line 3: rename file -> `kintransfer_file_provider_paths`

OrenZak: path prefix: `kintransfer`

transfer/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">transfer</string>
```
OrenZak, line 2: delete file ?

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+            } else {
+                startService(intent);
+            }
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private String createNotificationChannel() {
+        String channelId = CHANNEL_ID;
+        String channelName = CHANNEL_NAME;
```
OrenZak, line 155: two strings above can be removed

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+        NotificationChannel chan = new NotificationChannel(channelId, channelName, NotificationManager.IMPORTANCE_DEFAULT);
+        chan.setLightColor(Color.BLUE);
+        chan.setImportance(NotificationManager.IMPORTANCE_NONE);
+        chan.setLockscreenVisibility(Notification.VISIBILITY_PRIVATE);
+        NotificationManager service = getSystemService(NotificationManager.class);
+        if (service != null) {
```
OrenZak, line 161: what happens if service is null ?

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+        String authority = getPackageName() + "." + FILE_PROVIDER_NAME;
+        final Uri uri = FileProvider.getUriForFile(this, authority, file);
+        intent.setFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+        intent.putExtra(EXTRA_PACKAGE_KEY, getApplication().getPackageName());
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
+            revokeUriPermission(uri, Intent.FLAG_GRANT_READ_URI_PERMISSION);
+        }
+        intent.setDataAndType(uri, getContentResolver().getType(uri));
+        final List<ResolveInfo> services = getPackageManager().queryIntentServices(intent, 0);
+        if (!services.isEmpty()) {
```
OrenZak, line 143: what will happen if isEmpty ?

transfer/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
...
+# You can control the set of applied configuration files using the
+# proguardFiles setting in build.gradle.
+#
+# For more details, see
+#   http://developer.android.com/guide/developing/tools/proguard.html
+
+# If your project uses WebView with JS, uncomment the following
+# and specify the fully qualified class name to the JavaScript interface
+# class:
+#-keepclassmembers class fqcn.of.javascript.interface.for.webview {
```
YohayBar, line 11: if this file has not instruction better to delete it

transfer/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+        versionName "1.0"
+
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
```
YohayBar, line 21: remove this line:
1. in order for it be in use it needs to be declared in consumerProguardFiles.
2. We shouldn't enforce pro guard-android.txt rules on hosting app
3. proguard-rules.pro is empty

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                }
+
+                @Override
+                public void onAccountInfoFailed(String reason) {
+                }
+            });
+
+        }
+        stopForeground();
```
OrenZak, line 75: Can `stopForeground` be called before `onAccountInfoReady` ? let's say it takes time to be received.

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+    @Override
+    protected void onHandleIntent(final @Nullable Intent intent) {
+        startForeground();
+
+        if (intent != null && intent.getAction().equals(ACTION_KIN_TRANSFER_INFO) && intent.hasExtra(EXTRA_ACTION_KEY)
+                && intent.hasExtra(EXTRA_PACKAGE_KEY) && intent.hasExtra(EXTRA_SERVICE_FULL_PATH_KEY)) {
+            getAccountInfo(new AccountInfoListener() {
+                @Override
+                public void onAccountInfoReady(String data) {
+                    if (createFile(data)) {
```
OrenZak, line 62: else -> stopForeground ?

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+                        startServiceKinAccountReceiver(intent.getStringExtra(EXTRA_ACTION_KEY),
```
OrenZak, line 63: you can use `ACTION_KIN_TRANSFER_INFO` directly

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+    private static String FILE_PROVIDER_DIR_NAME = "kin_account_info";
+    private static String FILE_NAME = "kinAccountInfo.txt";
+    private static String ACTION_KIN_TRANSFER_INFO = "kin.transfer.info";
+    private static String EXTRA_PACKAGE_KEY = "EXTRA_PACKAGE_KEY";
+    private static String EXTRA_ACTION_KEY = "EXTRA_ACTION_KEY";
+    private static String EXTRA_SERVICE_FULL_PATH_KEY = "EXTRA_SERVICE_FULL_PATH_KEY";
+    private static String FILE_PROVIDER_NAME = "AccountInfoFileProvider";
+    private static String CHANNEL_NAME = "kin transfer";
+    private static String CHANNEL_ID = "0";
+    private static String NOTIFICATION_TRANSFER = "transfer kin";
```
OrenZak, line 41: separate strings with new line for those who are in the same category.

transfer/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,18 @@
...
+            android:grantUriPermissions="true"
+            android:name=".AccountInfoFileProvider">
+
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/file_provider_paths"/>
+        </provider>
+
+
+
```
OrenZak, line 16: remove lines

transfer/proguard-rules.pro
OrenZak, line 11: delete file ?

transfer/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+android {
+    compileSdkVersion 26
+
+
+
+    defaultConfig {
+        minSdkVersion 19
+        targetSdkVersion 26
+        versionCode 1
+        versionName "1.0"
```
OrenZak, line 12: look on `recovery` module and use same syntax, dependencies, buildTypes.

transfer/build.gradle
```diff
@@ -0,0 +1,34 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 26
+
+
+
+    defaultConfig {
+        minSdkVersion 19
```
YohayBar, line 9: because this module is part of the entire repo you can and should use 
 `minSdkVersion = rootProject.ext.minSdkVersion`

transfer/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+
+}
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+
+    implementation 'com.android.support:appcompat-v7:26.1.0'
```
YohayBar, line 30: the same, you can and should use `implementation devDependencies.appCompat_v7`

transfer/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
```
YohayBar, line 20: minifyEnabled false? you don't have to worry about it in this module , only if you have to add keep statement

OrenZak (changes requested):
see my comments

transfer/proguard-rules.pro
YohayBar: yes

transfer/src/test/java/com/kin/ecosystem/transfer/ExampleUnitTest.java
YohayBar: add test :)

app/build.gradle
YohayBar: I think this should move to kin-ecosystem-sdk if we want to enable it by default to all partners

shayb: i added it to sample app for testing and showing how to implement  it in your side.
this is up to you if needs to be included

app/src/main/java/com/ecosystem/kin/app/AccountInfoServiceImpl.java
shayb: because we wake service form the other side 
the name of the service is the same for all app
its the package that is changed
we can update clearer explanation

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
shayb: for now we might need only the public address 
later on we might need more info about the account
so i prefer to write something generic and not specific

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
shayb: might be removed lets talk about it

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
shayb: im not sure that this is the job of this module 
the one that provide the info need to check that its valid
if the length is fine but the address doesnt exist

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
shayb: why do you prefer that?

app/build.gradle
OrenZak: As we talked it can be in the SDK module

app/src/main/java/com/ecosystem/kin/app/AccountInfoServiceImpl.java
```diff
@@ -0,0 +1,64 @@
...
+*
+* */
+public class AccountInfoServiceImpl extends AccountInfoService {
+    public AccountInfoServiceImpl() {
+        super("AccountInfoService");
+    }
```
OrenZak, line 17: So only the service name should not be changed.
Can we have it in the SDK module? so the developer doesn't need to do anything.

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
YohayBar: not sure that using listener in intent service is considered best practice plus combining Intent service with foreground service. Why do we need  to show the user notification UI (as SDK we don't want to force notification on our hosting apps) - in the intent service we can simply call synchronous method that return the public key

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+        fileDirPath = getFilesDir().getAbsolutePath() + File.separator + FILE_PROVIDER_DIR_NAME;
+        fileFullPath = fileDirPath + File.separator + FILE_NAME;
+    }
+
+    private void stopForeground() {
+        stopForeground(true);
+        stopSelf();
+    }
+
+    private void startForeground() {
```
YohayBar, line 90: please remove this - we don't want to force any kind of notification (need design, approval from hosting app etc.)

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
OrenZak: So let's deal with it later, for now, you can just use `publicAddress`.
Also for something more generic, I would create an `AccountInfo` object and add params there.

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
OrenZak: 👍

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    public abstract void getAccountInfo(AccountInfoListener listener);
+
+    @Override
+    protected void onHandleIntent(final @Nullable Intent intent) {
+        startForeground();
+
+        if (intent != null && intent.getAction().equals(ACTION_KIN_TRANSFER_INFO) && intent.hasExtra(EXTRA_ACTION_KEY)
+                && intent.hasExtra(EXTRA_PACKAGE_KEY) && intent.hasExtra(EXTRA_SERVICE_FULL_PATH_KEY)) {
```
YohayBar, line 58: I suggest the first extract all keys into private members, check if not null and reuse later.

YohayBar: What will happen (in edge case) if interservice is restarted?

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+            });
+
+        }
+        stopForeground();
+    }
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        fileDirPath = getFilesDir().getAbsolutePath() + File.separator + FILE_PROVIDER_DIR_NAME;
```
YohayBar, line 81: place this fileDirPath and  fileFullPath in createFile so you could add unit testing

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoService.java
```diff
@@ -0,0 +1,167 @@
...
+            startForeground(1, notification);
+        } else {
+            NotificationCompat.Builder builder = new NotificationCompat.Builder(this);
+            builder.setContentTitle(getString(R.string.app_name)).setContentText(NOTIFICATION_TRANSFER).setPriority(NotificationCompat.PRIORITY_DEFAULT).setAutoCancel(true);
+            Notification notification = builder.build();
+            startForeground(1, notification);
+        }
+    }
+
+    private boolean createFile(String data) {
```
YohayBar, line 104: add unit testing

PR #244, opened 2018-12-31
move to v2 apis

OrenZak:
#### Main purpose:
Server updated new version for the apis
#### Technical description:
- update login flow


sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -240,50 +229,52 @@ public void run() {
 				}
 			});
 		}
-
-		String deviceID = AuthRepository.getInstance().getDeviceID();
```
YohayBar, line 244: we need to keep the device_id for backward compatibility and continues BI. How about use device_id and  app_device_id same as user_id and app_user_id?

OrenZak: not sure, because its an overhead to keep both. I'll talk with Oded about it.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -51,4 +52,14 @@ public int getAccountIndex() {
...
 		blockchainSharedPreferences.edit().putInt(ACCOUNT_INDEX_KEY, index).apply();
 	}
+
+	@Override
+	public void setAppId(@NonNull String appId) {
+		blockchainSharedPreferences.edit().putString(APP_ID_KEY, appId).apply();
+	}
+
+	@Override
+	public String getAppId() {
```
YohayBar: why do we need to store app_id in blockchain? - it is better to hold a single point of truth

YohayBar (approved):

PR #245, opened 2019-01-02
Enable Auto Backup and include required shared preference files

YohayBar:
#### Main purpose:
Enable auto backup in sample app to test and provide guidelines to DS 
#### Technical description:
Enable auto backup and set files that need to be included - all other files will be excluded
#### Dilemmas you faced with? 
Apps that disabled auto backup might not want to auto-backup all data - using include files solve this issue.
Also, this is not a full proof solution and it doesn't guarantee that wallet will be auto backup
#### Tests:
tested with https://developer.android.com/guide/topics/data/testingbackup and check that only included files were included in the backup.
#### Documentation:
Updated README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,28 @@ For Beta: ("beta")
 ```
 And for Production use: `@string/kinecosystem_environment_production` as value. 
 
+## Enable Auto backup ##
+Android Auto Backup automatically backs up apps data that target and run on Android 6.0 (API level 23) and above.
+Enabling auto backup will provide better experience when user uninstall and reinstall the app.
+No extra action is needed, if auto backup is enabled and ```android:fullBackupContent``` is not defined.
+Otherwise the app module AndroidManifest.xml should include these attributes.
+```xml
+ <application
+        android:allowBackup="true"
+        tools:replace="android:allowBackup"
+        android:fullBackupContent="@xml/backup_rules">  
+</application>
+```
+and ```backup_rules.xml``` should include 
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <include domain="sharedpref" path="KinKeyStore_kinecosystem_store.xml"/>
+</full-backup-content>
+```
+Auto backup is not 100% guaranteed and may not work for all users.
+To learn more on Auto Backup features and constraints see [Android Auto Backup](https://developer.android.com/guide/topics/data/autobackup) 
+
 
 >**NOTES:**
 >* When working with the Beta environment, you can only register up to 1000 users. An attempt to register additional users will result in an error.
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -6,8 +6,9 @@
     <uses-permission android:name="android.permission.INTERNET" />
 
     <application
-        android:allowBackup="false"
+        android:allowBackup="true"
         tools:replace="android:allowBackup"
+        android:fullBackupContent="@xml/backup_rules"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -24,6 +24,28 @@ For Beta: ("beta")
 ```
 And for Production use: `@string/kinecosystem_environment_production` as value. 
 
+## Enable Auto backup ##
+Android Auto Backup automatically backs up apps data that target and run on Android 6.0 (API level 23) and above.
+Enabling auto backup will provide better experience when user uninstall and reinstall the app.
+No extra action is needed, if auto backup is enabled and ```android:fullBackupContent``` is not defined.
+Otherwise the app module AndroidManifest.xml should include these attributes.
+```xml
+ <application
```
OrenZak, line 33: remove first space

README.md
```diff
@@ -24,6 +24,28 @@ For Beta: ("beta")
...
+        android:allowBackup="true"
+        tools:replace="android:allowBackup"
+        android:fullBackupContent="@xml/backup_rules">  
+</application>
+```
+and ```backup_rules.xml``` should include
```
OrenZak, line 39: should include the following lines:

README.md
```diff
@@ -24,6 +24,28 @@ For Beta: ("beta")
...
+        android:allowBackup="true"
+        tools:replace="android:allowBackup"
+        android:fullBackupContent="@xml/backup_rules">  
+</application>
+```
+and ```backup_rules.xml``` should include 
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <include domain="sharedpref" path="KinKeyStore_kinecosystem_store.xml"/>
```
OrenZak, line 43: Do we need only to keep this file?
Shouldn't we keep also blockchain source file ?

app/src/main/AndroidManifest.xml
```diff
@@ -6,8 +6,9 @@
     <uses-permission android:name="android.permission.INTERNET" />
 
     <application
-        android:allowBackup="false"
+        android:allowBackup="true"
         tools:replace="android:allowBackup"
+        android:fullBackupContent="@xml/backup_rules"
```
OrenZak, line 11: if we won't set it what will happen?

app/src/main/java/com/ecosystem/kin/app/MainActivity.java
```diff
@@ -177,6 +178,15 @@ public void onClick(View v) {
...
 			.setText(getString(R.string.version_name, BuildConfig.VERSION_NAME));
+
+		final TextView backupSessionsTextView = findViewById(R.id.backup_session);
+		final int numberOfBackupSessions = AutoBackupRepo.getNumberOfBackupedSessions(getApplicationContext());
+		backupSessionsTextView.setText(getString(R.string.backup_sessions, numberOfBackupSessions));
+
+		final TextView sessionsTextView = findViewById(R.id.sessions);
+		final int numberOfSessions = AutoBackupRepo.getNumberOfSessions(getApplicationContext());
+		sessionsTextView.setText(getString(R.string.sessions, numberOfSessions));
+		AutoBackupRepo.incrementNumberOfSessions(getApplicationContext());
```
OrenZak, line 189: Do we really need that?
Can't we do (unit, android)tests instead?

app/src/main/java/com/ecosystem/kin/app/model/AutoBackupRepo.java
```diff
@@ -0,0 +1,50 @@
+package com.ecosystem.kin.app.model;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.support.annotation.NonNull;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class AutoBackupRepo {
```
OrenZak, line 9: Same, do we really need that ? can't we do tests ?

PR #247, opened 2019-01-07
Eco 923/add login logout ui sample app

OrenZak:
#### Main purpose:
Add login, logout UI for sample app, so we can test login and logout flows.
#### Technical description:
- Added splash screen so we can decide if the user is logged in or not.
- Added login screen
- Added logout button on MainAcvitity - not working yet.


YohayBar (approved):
👍

PR #248, opened 2019-01-07
Shay/kin account info activity

shayb:
#### Main purpose:
Add ability to transfer account info - public address back to the caller application
I added AccountInfoActivity.java with some basic layout so you can see the usage of the object AccountInfoManager - the activity can be over written later on with the final design
#### Technical description:
a caller application start activity for result and open activity in the sdk core where the user need to approve and send back result. The result is a file uri with data - public address - of the user account
#### Dilemmas you faced with? 
(Not a must)
#### Tests:
Need to work on tests 
#### Documentation:
(Source/readme.md)


transfer/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,23 @@
...
+
+    <application>
+
+        <activity android:name=".AccountInfoActivity"
+            android:screenOrientation="portrait"
+            android:exported="true"
+            />
+
+        <provider
+            android:name=".KinTransferAccountInfoFileProvider"
```
berryve, line 7: name could be simplified to AccountInfoFileProvider

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoManager.java
```diff
@@ -0,0 +1,86 @@
...
+            listener.onError("Cant write to file " + e.getMessage());
+            return;
+        }
+        listener.onSuccess();
+    }
+
+    //Can be called anytime (before init complete), when user decline transfer
+    void respondCancel(@NonNull Activity activity) {
+        Intent intent = new Intent();
+        activity.setResult(Activity.RESULT_CANCELED, intent);
```
berryve, line 65: best to delete the file if the user cancels - but need to be careful about it and make sure it is recreated it if the activity is launched again.

berryve: or only create the file if the user approves...

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoManager.java
```diff
@@ -0,0 +1,86 @@
...
+
+    //Can be called only after init returned success, hence file is not null
+    void respondOk(@NonNull Activity activity) {
+        if (file != null) {
+            Intent intent = new Intent();
+            String authority = activity.getPackageName() + "." + FILE_PROVIDER_NAME;
+            final Uri uri = FileProvider.getUriForFile(activity, authority, file);
+            intent.setFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+            if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
+                activity.revokeUriPermission(uri, Intent.FLAG_GRANT_READ_URI_PERMISSION);
```
berryve, line 76: why is this needed ?

transfer/src/main/AndroidManifest.xml
shayb: this is changed to this becuase of Oren request so developer will not delete it by accident

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoManager.java
shayb: better to create before user approve cause it will delay the user if he press ok 
and then he needs to wait 
and what happen if it fails to create after send ok
better note the user before pressing something
not good idea to delete it cause other app needs to read it later on

transfer/src/main/java/com/kin/ecosystem/transfer/AccountInfoManager.java
shayb: the way file provider ask for grand read permission

berryve: doesn't make sense - revoking permission?

PR #250, opened 2019-01-09
Eco 775/mapping kin user to wallet address

OrenZak:
#### Main purpose:
Mapping kin user id to wallet address so we can load the relevant wallet for this user.
#### Technical description:
- Create a mapping from kin_user_id to wallet address
- Save the mapping in `BlockchainSource.Local` sharedpref
- load last active account or create an account if no matching found
#### Tests:
- Added test to `BlockchainSourceImplTest` regarding loading user wallet
- Refactor `BlockchainSourceImplTest` to Kotlin


app/src/main/java/com/ecosystem/kin/app/JwtUtil.java
```diff
@@ -18,208 +18,216 @@
...
+	public static String generateSpendOfferExampleJWT(String appID, String userID, String deviceId, String offerID) {
+		return getBasicJWT(appID)
+			.setSubject(JWT_SUBJECT_SPEND)
+			.claim(JWT_CLAIM_OBJECT_OFFER_PART, createOfferPartExampleObject(offerID))
+			.claim(JWT_CLAIM_OBJECT_SENDER_PART,
+				new JWTSenderPart(userID, deviceId, "Bought a sticker", "Lion sticker"))
+			.signWith(SignatureAlgorithm.RS512, getRS512PrivateKey()).compact();
+	}
+
+	public static String generateEarnOfferExampleJWT(String appID, String userID, String deviceId) {
```
YohayBar: should deviceId id be part of getBasicJWT?

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
```diff
@@ -90,7 +90,7 @@ public String getUserID() {
 
 	@Override
 	public String getEcosystemUserID() {
-		return localData.getEcosystemUserID();
+		return cachedAuthToken == null ? localData.getEcosystemUserID() : cachedAuthToken.getEcosystemUserID();
```
YohayBar, line 93: cached auth token complicate the logic localData.getEcosystemUserID() should handle caching internally if needed

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -81,7 +86,9 @@ private BlockchainSourceImpl(@NonNull EventLogger eventLogger, @NonNull final Ki
 		this.authRepository = authRepository;
 		this.kinClient = kinClient;
 		this.local = local;
-		this.activeAccountIndex = local.getAccountIndex();
+		Logger.log(new Log().withTag(TAG)
+			.put("BlockchainSourceImpl authRepository.getEcosystemUserID()", authRepository.getEcosystemUserID()));
+		this.currentUserId = authRepository.getEcosystemUserID();
```
YohayBar, line 90: same I prefer one point of truth if you want caching to improve performance it should be handled by authRepository.getEcosystemUserID()

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +108,81 @@ public static BlockchainSourceImpl getInstance() {
...
 	}
 
-	private void createKinAccountIfNeeded() throws BlockchainException {
-		if (kinClient.hasAccount()) {
+	/**
+	 * Support backward compatibility,
+	 * load the old account and migrate to current multiple users implementation.
+	 */
+	private void migrateToMultipleUsers(String kinUserId) {
+		final int accountIndex = local.getAccountIndex();
```
YohayBar: is there a risk of migration from older SDK versions before backup and restore that didn't have the concept of accountIndex?

YohayBar: we need to start thinking about standardize migration and keep track of migration history so we can make sure that we also support non-incremental updates

sdk/src/main/java/com/kin/ecosystem/base/KinEcosystemInitiator.java
```diff
@@ -19,7 +21,11 @@ protected static void init(Context context) {
 		try {
 			Kin.initialize(context);
 			// If we had process restart we should load the account.
-			BlockchainSourceImpl.getInstance().createAccount();
+			KinAccount account = BlockchainSourceImpl.getInstance().getKinAccount();
+			if (account == null) {
+				final String kinUserId = AuthRepository.getInstance().getEcosystemUserID();
+				BlockchainSourceImpl.getInstance().loadAccount(kinUserId);
```
YohayBar: why do we have 2 points that calls BlockchainSourceImpl.getInstance().loadAccount?

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -43,13 +49,38 @@ public void setBalance(int balance) {
...
+	public void setActiveUserWallet(String userId, String publicAddress) {
+		Set<String> currentWallets = getUserWallets(userId);
+		Set<String> updatedWallets = currentWallets != null ? new LinkedHashSet<>(currentWallets) : new LinkedHashSet<String>();
+		updatedWallets.remove(publicAddress); // Remove if exists
+		updatedWallets.add(publicAddress); // Add to the end
+		Editor editor = blockchainSharedPreferences.edit();
+		editor.putString(CURRENT_KIN_USER_ID, userId);
+		editor.putString(CURRENT_WALLET_ADDRESS, publicAddress);
+		editor.putStringSet(userId, updatedWallets).apply();
+		editor.apply();
```
YohayBar, line 76: apply is called twice

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +108,81 @@ public static BlockchainSourceImpl getInstance() {
...
-			} catch (CreateAccountException e) {
-				throw ErrorUtil.getBlockchainException(e);
-			}
+			account = createAccount();
+		}
+		local.setActiveUserWallet(kinUserId, account.getPublicAddress());
+	}
+
+	@Nullable
+	private String getLastWalletAddress(Set<String> wallets) {
```
YohayBar: getLastWalletAddress should be part of local and probably we can make getUserWallets private.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +108,81 @@ public static BlockchainSourceImpl getInstance() {
...
+		Set<String> wallets = local.getUserWallets(kinUserId);
+		if (kinClient.hasAccount() && wallets != null && wallets.size() > 0) {
+			Logger.log(new Log().withTag(TAG).text("createOrLoadAccount").put("currentUserId", currentUserId)
+				.put("kinUserId", kinUserId));
+			final String lastWalletAddress;
+			if (!StringUtil.isEmpty(currentUserId)) {
+				if (currentUserId.equals(kinUserId)) {
+					// Get last active wallet from cache
+					lastWalletAddress = local.getCurrentWalletAddress();
+				} else {
```
YohayBar: why do we need the if else we can simply getLastWalletAddress of kinUserId

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +108,81 @@ public static BlockchainSourceImpl getInstance() {
...
+					lastWalletAddress = getLastWalletAddress(wallets);
+					this.currentUserId = kinUserId;
+				}
+
+				// Match between last wallet address to wallets on device.
+				if (!StringUtil.isEmpty(lastWalletAddress)) {
+					for (int i = 0; i < kinClient.getAccountCount(); i++) {
+						KinAccount account = kinClient.getAccount(i);
+						if (lastWalletAddress.equals(account.getPublicAddress())) {
+							this.account = account;
```
YohayBar: not sure how the account is implemented but if it doesn't store the public key it needs to recalculate it from the private key which could be costly operation

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
YohayBar: I suggest to  seprate into several methods
local.getLastWalletAddress(userId).
If not existed create account 
if existed call a method getAccount(publicAddress)
if getAccount(publicAddress) doesn't return an account we have a problem

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -43,13 +49,38 @@ public void setBalance(int balance) {
...
+	@Nullable
+	@Override
+	public String getCurrentWalletAddress() {
+		return blockchainSharedPreferences.getString(CURRENT_WALLET_ADDRESS, null);
+	}
+
 	@Override
 	public int getAccountIndex() {
-		return blockchainSharedPreferences.getInt(ACCOUNT_INDEX_KEY, 0);
+		return blockchainSharedPreferences.contains(ACCOUNT_INDEX_KEY) ? blockchainSharedPreferences.getInt(ACCOUNT_INDEX_KEY, 0) : NOT_EXIST;
```
YohayBar: returning 0 will handle version before ACCOUNT_INDEX_KEY?

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: it's not for cache this is the logic of blockchainsource.

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
OrenZak: no, it's the repo responsibility and it's not complicated.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: there is no risk.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: I don't think it should be part of local, because it's the repo logic.
the local can supply the **current** last active wallet and getUserWallets nothing more.
to get the last from the user wallets is the repo responsibility.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: 👍

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: what do you mean by `doesn't store the public key`?

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
YohayBar: OK

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +107,70 @@ public static BlockchainSourceImpl getInstance() {
...
+	 */
+	private void migrateToMultipleUsers(String kinUserId) {
+		final boolean isMigrated = local.getIsMigrated();
+		if (!isMigrated && kinClient.hasAccount()) {
+			final int accountIndex = local.getAccountIndex();
+			KinAccount account;
+			if(accountIndex == NOT_EXIST) {
+				account = kinClient.getAccount(0);
+			} else {
+				Logger.log(new Log().withTag(TAG).put("migrateToMultipleUsers user:", kinUserId));
```
YohayBar: add logger also accountIndex == NOT_EXIST

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +107,70 @@ public static BlockchainSourceImpl getInstance() {
...
+	}
+
+	private void createOrLoadAccount(String kinUserId) throws BlockchainException {
+		final String lastWalletAddress = local.getLastWalletAddress(kinUserId);
+		if (kinClient.hasAccount() && !StringUtil.isEmpty(lastWalletAddress)) {
+			Logger.log(new Log().withTag(TAG).text("createOrLoadAccount").put("currentUserId", currentUserId)
+				.put("kinUserId", kinUserId));
+
+			// Match between last wallet address to wallets on device.
+			if (!StringUtil.isEmpty(lastWalletAddress)) {
```
YohayBar: you already checked that !StringUtil.isEmpty(lastWalletAddress)

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +107,70 @@ public static BlockchainSourceImpl getInstance() {
...
+						this.account = account;
+						break;
+					}
+				}
+			}
+
+			// No matching found
 			if (account == null) {
-				account = kinClient.getAccount(activeAccountIndex);
+				account = createAccount();
```
YohayBar, line 155: this should never happens if we have lastWalletAddress we must have a matching account.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +107,70 @@ public static BlockchainSourceImpl getInstance() {
...
-				account = kinClient.addAccount();
-				setAccountIndex(0);
-			} catch (CreateAccountException e) {
-				throw ErrorUtil.getBlockchainException(e);
-			}
+			account = createAccount();
+		}
+
+		currentUserId = kinUserId;
+		local.setActiveUserWallet(kinUserId, account.getPublicAddress());
```
YohayBar, line 163: in theory we can call  ` local.setActiveUserWallet(kinUserId, account.getPublicAddress());` only when calling 
    
 `account = createAccount();` otherwise it is already set

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -43,13 +49,50 @@ public void setBalance(int balance) {
...
+		editor.apply();
+	}
+
+	private LinkedHashSet<String> getUserWallets(String kinUserId) {
+		Set<String> wallets = blockchainSharedPreferences.getStringSet(kinUserId, null);
+		return wallets != null ? new LinkedHashSet<>(wallets) : new LinkedHashSet<String>();
+	}
+
+	@Override
+	public void deleteAccountIndexKey() {
```
YohayBar, line 85: removeAccountIndexKey

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -43,13 +49,50 @@ public void setBalance(int balance) {
...
+		currentWallets.remove(publicAddress); // Remove if exists
+		currentWallets.add(publicAddress); // Add to the end
+		Editor editor = blockchainSharedPreferences.edit();
+		editor.putString(CURRENT_KIN_USER_ID, kinUserId);
+		editor.putStringSet(kinUserId, currentWallets).apply();
+		editor.apply();
+	}
+
+	private LinkedHashSet<String> getUserWallets(String kinUserId) {
+		Set<String> wallets = blockchainSharedPreferences.getStringSet(kinUserId, null);
```
YohayBar, line 80: are you sure SharedPreferences doesn't alter the order of the set between set and get?

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: i would keep it there for safety

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
YohayBar: OK

YohayBar (approved):

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
OrenZak: I did a few tests and it looks ok, we can ask QA to verify.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
OrenZak: better to be on the safe side.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -101,24 +107,69 @@ public static BlockchainSourceImpl getInstance() {
...
+			KinAccount account;
+			if(accountIndex == NOT_EXIST) {
+				account = kinClient.getAccount(0);
+				Logger.log(new Log().withTag(TAG).put("migrateToMultipleUsers accountIndex == NOT_EXIST, kinUserId", kinUserId));
+			} else {
+				Logger.log(new Log().withTag(TAG).put("migrateToMultipleUsers accountIndex", accountIndex).put("kinUserId", kinUserId));
+				account = kinClient.getAccount(accountIndex);
+				local.deleteAccountIndexKey();
+			}
+			local.setDidMigrate();
```
YohayBar, line 133: check first  'if (!isMigrated)' and call `local.setDidMigrate();` otherwise first time user will not migrate for the first time but in the second time isMigrate will still be false but this time hasAccount will be true. In an edge case where User A log in for the first time logout and login back as user B they will share account.

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
YohayBar: better know we have a problem then overriding  users wallet

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
YohayBar: Ok

PR #252, opened 2019-01-14
Eco 778/add logout api

OrenZak:
#### Main purpose:
Allowing logout from a user
#### Technical description:
- `Kin.logout()` API which handles the logout for the current user.
- Remove relevant cached info from all the repos

This PR based on #250


OrenZak:
This PR based on #250, make sure to sync with the epic after the merge.

common/src/main/java/com/kin/ecosystem/common/ObservableData.java
```diff
@@ -117,4 +117,8 @@ private void onChanged() {
             observer.onChanged(value.get());
         }
     }
+
+    public void removeAllObservers() {
```
YohayBar, line 121: this is dangerous.  one observer may cancel to all other

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthLocalData.java
```diff
@@ -118,5 +107,17 @@ public AuthToken getAuthTokenSync() {
...
 		}
 	}
+
+	@Override
+	public void logout() {
+		Editor editor = signInSharedPreferences.edit();
+		editor.remove(JWT_KEY);
+		editor.remove(USER_ID_KEY);
+		editor.remove(ECOSYSTEM_USER_ID_KEY);
+		editor.remove(DEVICE_ID_KEY);
```
YohayBar: do we want to delete Device ID?

OrenZak: nop

common/src/main/java/com/kin/ecosystem/common/ObservableData.java
OrenZak: what do you suggest?

YohayBar: why do you want to remove observers? you can disable and enable the observer maybe?

OrenZak: well, as i see it now, we allowing the developer to add observers although the user is not logged in so we should not remove them.
We should remove only the`paymentObservers` because we shouldn't listen for payments if the account not logged in.

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
```diff
@@ -54,7 +56,29 @@ public static AuthRepository getInstance() {
 	@Override
 	public void setJWT(@NonNull String jwt) throws ClientException {
 		this.jwt = jwt;
-		localData.setJWT(jwt);
+		JwtBody jwtBody = getJwtBody(jwt);
+		String currentUserID = localData.getUserID();
+		if (!StringUtil.isEmpty(currentUserID) && !currentUserID.equals(jwtBody.getUserId())) {
+			// Update server
+			remoteData.logout(null);
```
YohayBar: should you also call
```
		cachedAuthToken = null;
		jwt = null;
		localData.logout();
```

YohayBar: i.e  call ` internalLogout()`

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
```diff
@@ -149,6 +173,28 @@ public void onFailure(ApiException exception) {
 		});
 	}
 
+	@Override
+	public void logout() {
+		remoteData.logout(new Callback<Void, ApiException>() {
+			@Override
+			public void onResponse(Void response) {
+				internalLogout();
```
YohayBar: why do wait for a response?

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -461,6 +468,15 @@ public boolean updateActiveAccount(int accountIndex) {
 		return false;
 	}
 
+	@Override
+	public void logout() {
+		removeRegistration(paymentRegistration);
```
YohayBar: should we also check in login if user_id changed? If it did change we can throw an error - enforcing developer to explicitly call login before. Or we can logout implicitly?

core/src/main/java/com/kin/ecosystem/core/data/offer/OfferRepository.java
```diff
@@ -165,4 +165,9 @@ public boolean shouldDismissOnTap(@NonNull String offerId) {
 		final Boolean shouldDismissOnTap = nativeOfferMap.get(offerId);
 		return shouldDismissOnTap == null ? false : shouldDismissOnTap;
 	}
+
+	@Override
+	public void logout() {
+		nativeOfferList.removeAll();
```
YohayBar: should we also clear cachedOfferList?

core/src/main/java/com/kin/ecosystem/core/data/settings/SettingsDataSourceLocal.java
```diff
@@ -8,21 +8,19 @@
...
 
 	public SettingsDataSourceLocal(@NonNull final Context context) {
 		this.settingsDataSourceSharedPref = context.getSharedPreferences(SETTINGS_PREF_NAME_FILE_KEY, Context.MODE_PRIVATE);
 	}
 
 	@Override
-	public void setIsBackedUp(boolean isBackedUp) {
-		settingsDataSourceSharedPref.edit().putBoolean(IS_BACKED_UP_KEY, isBackedUp).apply();
+	public void setIsBackedUp(final String publicAddress, boolean isBackedUp) {
+		settingsDataSourceSharedPref.edit().putBoolean(publicAddress, isBackedUp).apply();
```
YohayBar: In the future, we might want to save additional information in setting related to single public address, maybe keep  IS_BACKED_UP_KEY and append publicAddress to it as the key?

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
OrenZak: no need,`local.setJwt(...)` overrides the relevant data.

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
OrenZak: otherwise, jwt can be deleted and the logout call will fail.

core/src/main/java/com/kin/ecosystem/core/data/offer/OfferRepository.java
OrenZak: only `cachedOfferList` 🥇

core/src/main/java/com/kin/ecosystem/core/data/settings/SettingsDataSourceLocal.java
OrenZak: 👍

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
OrenZak: i'll change that so we won't wait.

common/src/main/java/com/kin/ecosystem/common/ObservableData.java
YohayBar: I'm not sure we need to remove observer - if we log out this observer should not omit any value

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
```diff
@@ -52,9 +54,32 @@ public static AuthRepository getInstance() {
 	}
 
 	@Override
-	public void setJWT(@NonNull String jwt) throws ClientException {
+	public void setJWT(@NonNull String jwt) throws ClientException, UserLoggedInException {
 		this.jwt = jwt;
-		localData.setJWT(jwt);
+		final JwtBody jwtBody = getJwtBody(jwt);
+		final String currentUserID = localData.getUserID();
+		localData.setJWT(jwtBody);
```
YohayBar: if throw new UserLoggedInException don't call localData.setJWT(jwtBody);
The error tells the Host app to first log out and only then login again

YohayBar: Also, we should call remote logout - we enforce the host app to do logout before login in with different user

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceImpl.java
```diff
@@ -461,6 +468,15 @@ public boolean updateActiveAccount(int accountIndex) {
...
+		removeRegistration(balanceRegistration);
```
YohayBar, line 474: you can also set paymentRegistration and balanceRegistration to null

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSource.java
```diff
@@ -117,7 +119,9 @@ void sendTransaction(@NonNull String publicAddress, @NonNull BigDecimal amount,
 
 		void setActiveUserWallet(String kinUserId, String publicAddress);
 
-		void deleteAccountIndexKey();
+		void removeAccountIndexKey();
+
+		void clearCachedBalance();
```
YohayBar: maybe call it log out if in the future there will be an additional properties that need to be cleared

OrenZak: it the future we will split if needed it doesn't matter now.

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
OrenZak: it's an internal exception not related to the DS.
so `Kin.java` will know there is already logged in user. and clear other cached data.

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -189,12 +191,13 @@ public static void login(@NonNull String jwt, KinCallback<Void> loginCallback) {
...
 
 	private static void internalLogin(@NonNull final String jwt, final KinCallback<Void> loginCallback) {
-
 		try {
 			checkInstanceNotNull();
 			AuthRepository.getInstance().setJWT(jwt);
 		} catch (final ClientException exception) {
 			sendLoginFailed(exception, loginCallback);
+		} catch (UserLoggedInException e) {
+			clearCachedData();
```
YohayBar: If setJWT through exception we  can't assume it succeeded in setting the JWT

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -52,40 +54,53 @@ public void setBalance(int balance) {
...
 	public int getAccountIndex() {
-		return blockchainSharedPreferences.contains(ACCOUNT_INDEX_KEY) ? blockchainSharedPreferences.getInt(ACCOUNT_INDEX_KEY, 0) : NOT_EXIST;
+		return blockchainSharedPreferences.contains(ACCOUNT_INDEX_KEY) ? blockchainSharedPreferences
+			.getInt(ACCOUNT_INDEX_KEY, 0) : NOT_EXIST;
 	}
 
 	@Override
 	public void setActiveUserWallet(String kinUserId, String publicAddress) {
-		LinkedHashSet<String> currentWallets = getUserWallets(kinUserId);
+		List<String> currentWallets = getUserWallets(kinUserId);
```
YohayBar: You can be explicit about using ArrayList<String>

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -52,40 +54,53 @@ public void setBalance(int balance) {
...
 		editor.apply();
 	}
 
-	private LinkedHashSet<String> getUserWallets(String kinUserId) {
-		Set<String> wallets = blockchainSharedPreferences.getStringSet(kinUserId, null);
-		return wallets != null ? new LinkedHashSet<>(wallets) : new LinkedHashSet<String>();
+	private String createLongString(List<String> currentWallets) {
+		StringBuilder walletsStringBuilder = new StringBuilder();
+		for (int i = 0; i < currentWallets.size(); i++) {
+			walletsStringBuilder.append(currentWallets.get(i)).append(STRING_WALLETS_DELIMITER);
```
YohayBar: the string will end with STRING_WALLETS_DELIMITER

YohayBar: sb.deleteCharAt(sb.length() - 1)

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
```diff
@@ -52,40 +54,53 @@ public void setBalance(int balance) {
 	@Nullable
 	@Override
 	public String getLastWalletAddress(final String kinUserId) {
-		Set<String> wallets = getUserWallets(kinUserId);
-		String publicAddress = null;
-		for (String wallet : wallets) {
-			publicAddress = wallet;
-		}
-		return publicAddress;
+		ArrayList<String> wallets = getUserWallets(kinUserId);
```
YohayBar, line 57: is this logic unit tested?

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -261,6 +264,29 @@ private static void checkInstanceNotNull() throws ClientException {
...
+		}
+	}
+
+	/**
+	 * Clear cached old internal data
+	 */
+	private static void clearCachedData() {
+		BlockchainSourceImpl.getInstance().logout();
+		OrderRepository.getInstance().logout();
```
YohayBar, line 287: OfferReposetiry?

sdk/src/main/java/com/kin/ecosystem/main/presenter/EcosystemPresenter.java
```diff
@@ -43,6 +47,11 @@ public EcosystemPresenter(@NonNull IEcosystemView view, @NonNull SettingsDataSou
 		this.blockchainSource = blockchainSource;
 		this.navigator = navigator;
 		this.currentBalance = blockchainSource.getBalance();
+		try {
+			this.publicAddress = blockchainSource.getPublicAddress();
+		} catch (ClientException | BlockchainException e) {
+			// no-op, should not happen
```
YohayBar: add log just in case

sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.java
```diff
@@ -51,6 +55,11 @@ public SettingsPresenter(@NonNull final ISettingsView view, @NonNull final Setti
 		this.eventLogger = eventLogger;
 		this.accountManager = accountManager;
 		this.currentBalance = blockchainSource.getBalance();
+		try {
+			this.publicAddress = blockchainSource.getPublicAddress();
+		} catch (ClientException | BlockchainException e) {
+			// no-op, should not happen
```
YohayBar: add log just in case

YohayBar (approved):
see my comments otherwise 👍

core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceLocal.java
OrenZak: yes

sdk/src/main/java/com/kin/ecosystem/main/presenter/EcosystemPresenter.java
OrenZak: 👍

PR #253, opened 2019-01-15
ECO-976 - wrong x_sdk_version header value

OrenZak:
#### Main purpose:
SDK header value was wrong
#### Technical description:
- Use one version to all SDK related modules


gradle.properties
```diff
@@ -20,8 +20,4 @@ SAMPLE_VERSION_NAME=0.0.77
 
 KIN_ECOSYSTEM_SDK_VERSION_NAME=0.3.1
 
-KIN_ECOSYSTEM_CORE_VERSION_NAME=0.0.2
-
-KIN_ECOSYSTEM_COMMON_VERSION_NAME=0.0.2
-
 KIN_ECOSYSTEM_BACKUP_VERSION_NAME=0.0.2
```
YohayBar, line 23: why KIN_ECOSYSTEM_BACKUP_VERSION_NAME is different? what was the problem? versionName was overridden by one of the sub modules?

OrenZak: `ConfigurationImpl` is in `core` module so it was using the 0.0.2 in it's `build.gradle`

backup is different because it a separated module not related to the SDK itself and not affecting the SDK version.

YohayBar (approved):

PR #255, opened 2019-01-20
add bi events

OrenZak:
#### Main purpose:
Add BI events -> login, logout

### DIFF
--- a/core/src/main/java/com/kin/ecosystem/core/bi/events/UserLoginFailed.java
+++ b/core/src/main/java/com/kin/ecosystem/core/bi/events/UserLoginFailed.java
@@ -1,0 +1,181 @@
+
+package com.kin.ecosystem.core.bi.events;
+
+// Augmented by script
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+import com.kin.ecosystem.core.bi.Event;
+import com.kin.ecosystem.core.bi.EventsStore;
+
+
+/**
+ * User login failed
+ * 
+ */
+public class UserLoginFailed implements Event {
+    public static final String EVENT_NAME = "user_login_failed";
+    public static final String EVENT_TYPE = "business";
+
+    // Augmented by script
+    public static UserLoginFailed create() {
+        return new UserLoginFailed(
+            (Common) EventsStore.common(),
+            (User) EventsStore.user(),
+            (Client) EventsStore.client());
+    }
+
+    /**
+     * 
+     * (Required)
+     * 
+     */
+    @SerializedName("event_name")
+    @Expose
+    private String eventName = EVENT_NAME;
+    /**
+     * 
+     * (Required)
+     * 
+     */
+    @SerializedName("event_type")
+    @Expose
+    private String eventType = EVENT_TYPE;
+    /**
+     * common properties for all events
+     * (Required)
+     * 
+     */
+    @SerializedName("common")
+    @Expose
+    private Common common;
+    /**
+     * common user properties
+     * (Required)
+     * 
+     */
+    @SerializedName("user")
+    @Expose
+    private User user;
+    /**
+     * common properties for all client events
+     * (Required)
+     * 
+     */
+    @SerializedName("client")
+    @Expose
+    private Client client;
+
+    /**
+     * No args constructor for use in serialization
+     * 
+     */
+    public UserLoginFailed() {
+    }
+
+    /**
+     * 
+     * @param common
+
+     * @param client
+
+     * @param user
+     */
+    public UserLoginFailed(Common common, User user, Client client) {
+        super();
+        this.common = common;
+        this.user = user;
+        this.client = client;
+    }
+
+    /**
+     * 
+     * (Required)
+     * 
+     */
+    public String getEventName() {
+        return eventName;
+    }
+
+    /**
+     * 
+     * (Required)
+     * 
+     */
+    public void setEventName(String eventName) {
+        this.eventName = eventName;
+    }
+
+    /**
+     * 
+     * (Required)
+     * 
+     */
+    public String getEventType() {
+        return eventType;
+    }
+
+    /**
+     * 
+     * (Required)
+     * 
+     */
+    public void setEventType(String eventType) {
+        this.eventType = eventType;
+    }
+
+    /**
+     * common properties for all events
+     * (Required)
+     * 
+     */
+    public Common getCommon() {
+        return common;
+    }
+
+    /**
+     * common properties for all events
+     * (Required)
+     * 
+     */
+    public void setCommon(Common common) {
+        this.common = common;
+    }
+
+    /**
+     * common user properties
+     * (Required)
+     * 
+     */
+    public User getUser() {
+        return user;
+    }
+
+    /**
+     * common user properties
+     * (Required)
+     * 
+     */
+    public void setUser(User user) {
+        this.user = user;
+    }
+
+    /**
+     * common properties for all client events
+     * (Required)
+     * 
+     */
+    public Client getClient() {
+        return client;
+    }
+
+    /**
+     * common properties for all client events
+     * (Required)
+     * 
+     */
+    public void setClient(Client client) {
+        this.client = client;
+    }
+
+}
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/kin/ecosystem/core/bi/events/UserLoginFailed.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * User login failed
+ * 
+ */
+public class UserLoginFailed implements Event {
+    public static final String EVENT_NAME = "user_login_failed";
+    public static final String EVENT_TYPE = "business";
+
+    // Augmented by script
+    public static UserLoginFailed create() {
```
YohayBar: there should be an errorReason in LoginFailed - I asked BI to fix it so you will probably need to regenerate this file

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -200,6 +203,7 @@ private static void internalLogin(@NonNull final String jwt, final KinCallback<V
 			sendLoginFailed(exception, loginCallback);
 		}
 
+		eventLogger.send(UserLogoutRequested.create());
```
YohayBar: UserLogoutRequested?

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -215,6 +219,7 @@ public void onResponse(AuthToken authToken) {
 				AuthRepository.getInstance().updateWalletAddress(publicAddress, new KinCallback<Boolean>() {
 					@Override
 					public void onResponse(Boolean response) {
+						eventLogger.send(UserLoginSucceeded.create());
```
YohayBar: Need to talk with BI - we planned to only send login  if user currently logout or login as a different user - "when a user login with a different device /user_id , and this device is currently logout"

OrenZak: done already.

sdk/src/main/java/com/kin/ecosystem/Kin.java
OrenZak: refresh

YohayBar: I suggest changing isSameUser to something like getLoginStatus(jwt) which returns  IntDef of logut|sameUser|diffrentUser

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -192,43 +197,63 @@ public static void login(@NonNull String jwt, KinCallback<Void> loginCallback) {
...
-				});
+
+				// check server is synced with client, if synced -> skip updateWalletAddress
+				if (publicAddress != null && !publicAddress.equals(accountInfo.getUser().getCurrentWallet())) {
+					AuthRepository.getInstance().updateWalletAddress(publicAddress, new KinCallback<Boolean>() {
+						@Override
+						public void onResponse(Boolean response) {
+							eventLogger.send(UserLoginSucceeded.create());
+							isAccountLoggedIn = true;
+							instance.executorsUtil.mainThread().execute(new Runnable() {
```
YohayBar:
```
 isAccountLoggedIn = true;
							}								instance.executorsUtil.mainThread().execute(new Runnable() {
						});									@Override
					}									public void run() {

 									loginCallback.onResponse(null);
					@Override									}
					public void onFailure(KinEcosystemException exception) {								});
```
is used twice consider having a private method

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
```diff
@@ -222,28 +231,34 @@ private void setAccountInfo(AccountInfo accountInfo) {
...
+		if (cachedAccountInfo == null || isAuthTokenExpired(cachedAuthToken)) {
+			if (StringUtil.isEmpty(jwt)) {
+				if (callback != null) {
+					callback.onFailure(ErrorUtil.getClientException(ClientException.ACCOUNT_NOT_LOGGED_IN, null));
 				}
+			} else {
+				remoteData.getAccountInfo(new JWT(jwt), new Callback<AccountInfo, ApiException>() {
+					@Override
+					public void onResponse(AccountInfo accountInfo) {
+						setAccountInfo(accountInfo);
```
YohayBar, line 245: set account info should also set cachedAccountInfo

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -192,43 +197,63 @@ public static void login(@NonNull String jwt, KinCallback<Void> loginCallback) {
 	private static void internalLogin(@NonNull final String jwt, final KinCallback<Void> loginCallback) {
 		try {
 			checkInstanceNotNull();
-			if(!AuthRepository.getInstance().isSameUser(jwt)) {
-				logout();
+			@UserLoginState int loginState = AuthRepository.getInstance().getUserLoginState(jwt);
```
YohayBar: internal login can be called more than once - consider setting isAccountLoggedIn = false in the beginning of this method

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -192,43 +197,63 @@ public static void login(@NonNull String jwt, KinCallback<Void> loginCallback) {
...
-						});
-					}
-
-					@Override
-					public void onFailure(KinEcosystemException exception) {
-						sendLoginFailed(exception, loginCallback);
-					}
-				});
+
+				// check server is synced with client, if synced -> skip updateWalletAddress
```
YohayBar: consider calling ```AccountManagerImpl.getInstance().start();``` anywhere isAccountLoggedIn is set to true. if it set to false the SDK can't be used anyway. No need to start AccountManagerImpl is not login and it better to start it after we know the user can be login

sdk/src/main/java/com/kin/ecosystem/Kin.java
YohayBar: consider also changing isAccountLoggedIn to atomic boolean

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -248,6 +274,10 @@ public void run() {
 		});
 	}
 
+	private static String getMessage(KinEcosystemException exception) {
+		return exception.getCause() != null ? exception.getCause().getMessage() : exception.getMessage();
```
YohayBar, line 281: check exception is not null for safty

sdk/src/main/java/com/kin/ecosystem/Kin.java
YohayBar: set isAccountLoggedIn = false after checking loginstate and only if new or switched user

sdk/src/main/java/com/kin/ecosystem/Kin.java
OrenZak: The only place we changing to true is here after getting token

sdk/src/main/java/com/kin/ecosystem/Kin.java
OrenZak: the exception is never null otherwise, you can't throw an exception.

sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -192,43 +198,51 @@ public static void login(@NonNull String jwt, KinCallback<Void> loginCallback) {
...
 				String publicAddress;
 				try {
-					BlockchainSourceImpl.getInstance().loadAccount(authToken.getEcosystemUserID());
+					BlockchainSourceImpl.getInstance().loadAccount(accountInfo.getAuthToken().getEcosystemUserID());
 					publicAddress = BlockchainSourceImpl.getInstance().getPublicAddress();
 				} catch (final BlockchainException exception) {
 					sendLoginFailed(exception, loginCallback);
 					return;
 				}
 				AccountManagerImpl.getInstance().start();
```
YohayBar: I think that there is no need to call `AccountManagerImpl.getInstance().start();` incase of failure (it will listen to blockchain for no reason) so it better to place it in sendLoginSucceed

YohayBar (approved):
see last comment otherwise 👍

PR #256, opened 2019-01-21
update login and logout docs

OrenZak:
#### Documentation:
- Update all jwt requierments (device_id was added)
- Added logout section to `docs/CREATE_ACCOUNT.md`

### DIFF
--- a/docs/CREATE_ACCOUNT.md
+++ b/docs/CREATE_ACCOUNT.md
@@ -5,6 +5,7 @@ If your app presents Kin Spend and Earn offers to your users, then each user nee
 >**NOTE:** Kin Ecosystem SDK must be initialized before any interaction with the SDK, in order to do that you should call `Kin.initialize(…)` first.
 
 
+#### Login
 *To create or access a user’s Kin account:*
 
 Call `Kin.login(…)`, passing a JWT credentials.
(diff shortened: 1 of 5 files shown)


docs/CREATE_ACCOUNT.md
```diff
@@ -31,4 +32,19 @@ If that user already has a Kin account, the function only accesses the existing
...
     }
-```   
\ No newline at end of file
+```
+
+#### Logout
+*To release access from a user’s Kin account or switch account:*
+
+Call `Kin.logout()`, this is a synchronous call means that you can call `Kin.login(…)` immediately (for switching between users).
+After calling `logout` method, you should login again inorder to interact with the SDK API's, as Kin account is required.
```
nitzantomer: It should be:
> After calling the `logout` method, ...
Or
> After calling `logout`, ...

Also, `inorder` isn't right here, should be `in order`

docs/PEER_TO_PEER.md
```diff
@@ -34,6 +34,7 @@ Your app displays the offer, request user approval, and then [requests payment u
     },
     sender: {
         user_id: string; // optional: user_id who will perform the order
```
nitzantomer: no longer optional

docs/NATIVE_SPEND.md
```diff
@@ -37,6 +37,7 @@ A custom Spend offer allows your users to unlock unique spend opportunities that
 
    sender: {
           user_id: string; // optional: ID of purchasing user
```
nitzantomer: no longer optional

nitzantomer (changes requested):

nitzantomer (approved):

PR #262, opened 2019-02-05
add backup failed event

OrenZak:
#### Main purpose:
Bi event for backup failed
#### Technical description:
- send the event with error reason

### DIFF
--- a/sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.java
+++ b/sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.java
@@ -188,7 +188,8 @@ public class SettingsPresenter extends BasePresenter<ISettingsView> implements I
 	}
 
 	private String getErrorMessage(Throwable throwable) {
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.java
```diff
@@ -185,6 +187,10 @@ public void onFailure(Throwable throwable) {
 		});
 	}
 
+	private String getErrorMessage(Throwable throwable) {
+		return throwable.getCause() != null ? throwable.getCause().getMessage() : throwable.getMessage();
```
doodyparizada: could throwable be `null`?

doodyparizada (approved):
looks good for now

sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.java
OrenZak: should not be in that case, but I'll check anyway.

PR #269, opened 2019-02-17
Eco 1045/integrate migration module

OrenZak:
#### Main purpose:
Integrate with the migration module in order to check for future potential bugs/requests.
Old kin blockchain should work as usual.
#### Technical description:
- Replace `kin.core` lib with `kin-migration-module-android` lib.
- Change all objects accordingly from old into the new one.
- Update the relevant tests that were changed due to the lib replacement.
#### Tests:
- Were updated


sdk/src/main/java/com/kin/ecosystem/KinContext.kt
```diff
@@ -0,0 +1,6 @@
+package com.kin.ecosystem
+
+import android.content.Context
+import android.content.ContextWrapper
+
+class KinContext(context: Context): ContextWrapper(context.applicationContext)
```
yosriz, line 6: why is that? to avoid warning?

yosriz (approved):
LGTM, no critical code implemented yet here, that related to migration

sdk/src/main/java/com/kin/ecosystem/KinContext.kt
OrenZak: avoid the warning and having the context in `Kin.java` level.

PR #278, opened 2019-03-19
Eco 1149/support dark mode

OrenZak:
#### Main purpose:
Allow the developer to choose between light/ dark theme
#### Technical description:
- Added `KinTheme`
- Create two styles with different attrs values according to the theme name. The attrs are the difference between the light and dark modes.


core/src/main/java/com/kin/ecosystem/core/data/internal/ConfigurationLocalImpl.kt
```diff
@@ -0,0 +1,35 @@
...
+    private val configurationSharedPref: SharedPreferences
+
+    override var kinTheme: KinTheme?
+        get() = KinTheme.valueOf(configurationSharedPref.getString(KIN_THEME_KEY, KinTheme.LIGHT.name))
+        set(value) {
+            with(configurationSharedPref.edit()) {
+                kinTheme?.let { currentTheme ->
+                    if(currentTheme != value ) {
+                        value?.let { value ->
+                            putString(KIN_THEME_KEY, value.name).apply()
```
shayb, line 18: no defualt value?

shayb (approved):
look at my comment

sdk/src/main/res/drawable/kinecosystem_ic_arrow_right.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
```
shayb, line 6: are you sue vectors supported ?? did you check on device with minimum sdk

OrenZak: yup works check on API 19:
ref:  https://developer.android.com/guide/topics/graphics/vector-drawable-resources

core/src/main/java/com/kin/ecosystem/core/data/internal/ConfigurationLocalImpl.kt
OrenZak: there is, check the `get():` property

PR #286, opened 2019-04-03
Eco 1161/settings screen

OrenZak:
#### Main purpose:
Adjust settings screen to the new UI
#### Technical description:
- Refactor from activity to fragment
- Handled Dark and light mode


sdk/src/main/java/com/kin/ecosystem/ViewExt.kt
```diff
@@ -28,4 +28,17 @@ inline fun <T : ValueAnimator> T.withEndAction(crossinline action: T.() -> Unit)
             action()
         }
     })
+}
+
+inline fun <T : ValueAnimator> T.withActions(crossinline startAction: T.() -> Unit, crossinline endAction: T.() -> Unit) {
```
yosriz: i'm not 100% sure, but I think you can return `T` instead of `Unit`, and then you can act on a specific object without specifying it (like `when` operator)
so you're code in
`withActions(startAction = {
                    earnOrderRecyclerView.visibility = VISIBLE
                }, endAction = {
                    spendOrderRecyclerView.visibility = GONE
                })
`
can become:
`
spendOrderRecyclerView.withActions(startAction = {
                   visibility = VISIBLE
                }, endAction = {
                   visibility = GONE
                })
`

sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    private fun addBalanceObserver() {
+        balanceObserver = object : Observer<Balance>() {
+            override fun onChanged(value: Balance) {
+                currentBalance = value
+                if (isGreaterThenZero(value)) {
+                    updateSettingsIcon()
+                }
+            }
+        }.also {
```
yosriz, line 61: isn't it shorter without the `also`, or am I missing something?

sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.kt
```diff
@@ -0,0 +1,182 @@
...
+            blockchainSource.removeBalanceObserver(it, false)
+            balanceObserver = null
+        }
+    }
+
+    override fun backupClicked() {
+        try {
+            backupManager.backupFlow()
+        } catch (e: ClientException) {
+            // Could not happen in this case.
```
yosriz, line 98: until it will...

yosriz (approved):
LGTM

sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.kt
OrenZak: more readable like that from my point of view.

sdk/src/main/java/com/kin/ecosystem/settings/presenter/SettingsPresenter.kt
OrenZak: nah

PR #293, opened 2019-04-10
Nitzan/eco 1089/bi migration events

nitzantomer:
updated bi events


core/src/main/java/com/kin/ecosystem/core/data/blockchain/BlockchainSourceRemote.java
```diff
@@ -56,6 +73,7 @@ public void run() {
 
 				@Override
 				public void onSuccess(final String result, final int statusCode, final Map<String, List<String>> responseHeaders) {
+					eventLogger.send(MigrationBCVersionCheckSucceeded.create(getPublicAddress(), MigrationBCVersionCheckSucceeded.BlockchainVersion.fromValue(result)));
```
OrenZak: `MigrationBCVersionCheckSucceeded.BlockchainVersion.fromValue(result)`
this can cause a crash if the result not aligned with the Enum values

OrenZak (approved):
just see my comments

PR #299, opened 2019-04-15
Eco 1284/new ui bi

OrenZak:
#### Main purpose:
Add BI events
#### Technical description:
- Added new events
- Removed unused events


sdk/src/main/java/com/kin/ecosystem/marketplace/presenter/MarketplacePresenter.kt
```diff
@@ -47,6 +48,7 @@ class MarketplacePresenter(private val offerRepository: OfferDataSource,
     private var currentBalance: Balance = blockchainSource.balance
     private var publicAddress: String = blockchainSource.publicAddress
 
+    private var isPageOfferEventSent = AtomicBoolean(false)
```
shayb, line 51: why need atomic here?

OrenZak: it can be called from `getOffers()` which is async call, and also from `getCachedOffers()`

shayb (approved):

Issue #304, opened 2019-04-26
NullPointerException in EcosystemActivity


tangjingyuan:
Hi @OrenZak , a NPE occurred in EcosystemActivity. The SDK version is 1.0.1. Here is the crash log:
![image](https://user-images.githubusercontent.com/12523498/56780570-5ec24200-6812-11e9-94ef-1862ce1082f8.png)

OrenZak:
Hi @tangjingyuan thank you for letting me know, I will investigate it.

PR #305, opened 2019-04-28
move setTheme to EcosystemBaseActivity

OrenZak:
#### Main purpose:
- Application crashed while application restart by the OS.
- Solves #304
#### Technical description:
- Moved up `KinEcosystemInitiator` to it will be triggered before getting the Theme
- Moved setThem to `EcosystemBaseActivity` after `super.onCreate` and before `etContentView`


yosriz (approved):
✌️ so long crashes, you are not welcome here!

yosriz:
@OrenZak there's no problem for setting the theme before setting the layout, right?

OrenZak:
> @OrenZak here's no problem for setting the theme before setting the layout, right?

@yosriz No, we must do it before otherwise, it will crash.

PR #307, opened 2019-05-01
Eco 1359/uninitialized property onboarding

### LINKED ISSUE #303: kotlin.UninitializedPropertyAccessException in OnboardingFragment
Hi @OrenZak , we watched a kotlin exception occurred in OnboardingFragment. SDK version is 1.0.1. Here is the crash log:

image

OrenZak:
#### Main purpose:
Initialize navigator and set to relevant fragment after process restart
#### Technical description:
- Init migration + account logged in after process restart
- Set navigator to the visible fragment upon `EcosystemActivity.onStart()`
- Create `KinEcosystemBaseFragment` to support the above.
#### Tests:
Manual testing with process restart.

Closes #303


yosriz (approved):
seems legit to me 👍 
minor pedant comment, I would do a dedicate method for internal initialize, instead of deriving it from the class type. 🤓

PR #309, opened 2019-05-13
add all native offers and keep order

OrenZak:
#### Main purpose:
All the developer to control over the order of his offers.
Add new API - the ability to add a few offers at once.
#### Technical description:
- Create new API: `Kin.addAllNativeOffers(...)`
- Simplify the offers order: native offers first and MP offers after.
- Update sample app with the new API 
- Added unit test for those cases.
#### Tests:
- Convert tests to Kotlin :)
- Test the order of the offers
#### Documentation:
- On `Kin.java` regarding the new API
- Will be added to the next docs version (website)


app/src/main/java/com/ecosystem/kin/app/main/MainActivity.java
```diff
@@ -291,12 +303,52 @@ private void launchExperience(@EcosystemExperience final int experience) {
...
 		nativeSpendOffer = getNativeSpendOffer();
-		addNativeOffer(nativeSpendOffer, true);
 		nativeEarnOffer = getNativeEarnOffer();
-		addNativeOffer(nativeEarnOffer, false);
+		List<NativeOffer> nativeOfferList = new ArrayList<>(2);
+		Collections.addAll(nativeOfferList, nativeSpendOffer, nativeEarnOffer);
+		try {
+			//Add offers to the top
+			if (Kin.addAllNativeOffers(nativeOfferList)) {
+				Kin.addAllNativeOffers(nativeOfferList);
```
drEfrati: why this line is not in a if statement if the first one is?

core/src/main/java/com/kin/ecosystem/core/data/offer/OfferRepository.java
```diff
@@ -178,26 +127,30 @@ public void removeNativeOfferClickedObserver(@NonNull Observer<NativeOfferClickE
...
 			}
 			return false;
 		}
 		return false;
 	}
 
+	@Override
+	public boolean addAllNativeOffers(List<NativeOffer> nativeOfferList) {
+		boolean result = true;
+		for (int i = nativeOfferList.size() - 1; i >= 0 ; i--) {
```
drEfrati: just from curiosity, why do you add them in the opposite order? meaning from the last offer to the first one?

drEfrati (approved):
Looks good to me.
Just two little comments.

app/src/main/java/com/ecosystem/kin/app/main/MainActivity.java
OrenZak: should be removed thanks.

core/src/main/java/com/kin/ecosystem/core/data/offer/OfferRepository.java
OrenZak: because `addNativeOffer` adds the offer to index 0 (at the top) so I want to keep the order of the list.

PR #315, opened 2019-05-30
fix blank errors while login, native spend and earn

OrenZak:
#### Main purpose:
Reduce the amount of black error reasons on login_failed and as well for native_earn/spend_failed events.
#### Technical description:
- Aligned all the exceptions to be `KinEcosystemException` and use `ErrorUtil` to get it.
- Refactored abit `CreateExternalOrderCall` so it will be more clear to read.
- Moved some of the logic to `ExternalEarnOrderCall` and `ExternalSpendOrderCall`
- Added unit test for those cases(only happy flow meantime)
#### Tests:
 - Added unit test for `ExternalEarnOrderCall` and `ExternalSpendOrderCall`

### DIFF
--- a/core/src/main/java/com/kin/ecosystem/core/data/order/CreateExternalOrderCall.java
+++ b/core/src/main/java/com/kin/ecosystem/core/data/order/CreateExternalOrderCall.java
@@ -7,14 +7,8 @@ import com.kin.ecosystem.common.exception.KinEcosystemException;
 import com.kin.ecosystem.common.exception.ServiceException;
 import com.kin.ecosystem.common.model.Balance;
 import com.kin.ecosystem.core.bi.EventLogger;
-import com.kin.ecosystem.core.bi.events.EarnOrderCompletionSubmitted;
-import com.kin.ecosystem.core.bi.events.EarnOrderCreationFailed;
-import com.kin.ecosystem.core.bi.events.EarnOrderCreationReceived;
-import com.kin.ecosystem.core.bi.events.SpendOrderCompletionSubmitted;
-import com.kin.ecosystem.core.bi.events.SpendOrderCreationFailed;
-import com.kin.ecosystem.core.bi.events.SpendOrderCreationReceived;
+import com.kin.ecosystem.core.bi.events.GeneralEcosystemSdkError;
 import com.kin.ecosystem.core.data.blockchain.BlockchainSource;
-import com.kin.ecosystem.core.data.blockchain.BlockchainSource.SignTransactionListener;
 import com.kin.ecosystem.core.data.blockchain.Payment;
 import com.kin.ecosystem.core.network.ApiException;
 import com.kin.ecosystem.core.network.model.JWTBodyPaymentConfirmationResult;
@@ -23,6 +17,8 @@ import com.kin.ecosystem.core.network.model.OpenOrder;
 import com.kin.ecosystem.core.network.model.Order;
 import com.kin.ecosystem.core.util.ErrorUtil;
 import com.kin.ecosystem.core.util.ExecutorsUtil.MainThreadExecutor;
+import com.kin.ecosystem.core.util.JwtDecoder;
+import com.kin.ecosystem.core.util.StringUtil;
 import java.math.BigDecimal;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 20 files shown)


core/src/main/java/com/kin/ecosystem/core/data/order/CreateExternalOrderCall.java
```diff
@@ -114,8 +91,17 @@ public void run() {
 				String orderID = extractOrderID(e.getResponseHeaders());
 				getOrder(orderID);
 			} else {
-				sendOrderCreationFailedEvent(openOrder, e);
-				onOrderFailed(ErrorUtil.fromApiException(e));
+				OfferJwtBody offerJwtBody = null;
+				try {
+					offerJwtBody = JwtDecoder.getOfferJwtBody(orderJwt);
```
yosriz, line 96: please test this new branch

OrenZak: added unit test thanks.

core/src/test/kotlin/com/kin/ecosystem/core/data/order/ExternalEarnOrderCallTest.kt
```diff
@@ -0,0 +1,107 @@
...
+		on { recipientAddress } doAnswer { "some_address" }
+	}
+
+	private val order: Order = mock {
+		on { orderId } doAnswer { orderId }
+		on { offerId } doAnswer { offerId }
+		on { result } doAnswer { jwtBodyPaymentConfirmationResult }
+	}
+
+	private val payment: Payment = mock {
```
yosriz, line 54: why do you need to mock model pojos? (`Order`,`BlockchainData`)
I think it's better to simply instantiate them with the right values

core/src/test/kotlin/com/kin/ecosystem/core/data/order/ExternalEarnOrderCallTest.kt
```diff
@@ -0,0 +1,107 @@
...
+			verify(blockchainSource).addPaymentObservable(paymentCaptor.capture())
+			verify(blockchainSource).blockchainVersion
+			argumentCaptor<KinCallback<Order>>().apply {
+				verify(orderDataSource).submitEarnOrder(any(), isNull(), any(), capture())
+				firstValue.onResponse(order)
+			}
+			verify(eventLogger).send(any<EarnOrderCompletionSubmitted>())
+			paymentCaptor.firstValue.onChanged(payment)
+			verify(blockchainSource).removePaymentObserver(any())
+			argumentCaptor<KinCallback<Order>>().apply {
```
yosriz, line 90: can you refactor everything to be sync? I think it will be more readable and testable this way

core/src/test/kotlin/com/kin/ecosystem/core/data/order/ExternalEarnOrderCallTest.kt
```diff
@@ -0,0 +1,107 @@
...
+	private lateinit var externalEarnOrderCall: ExternalEarnOrderCall
+
+	@Before
+	override fun setUp() {
+		super.setUp()
+		externalEarnOrderCall = ExternalEarnOrderCall(orderDataSource, blockchainSource, orderJwt, eventLogger, externalOrderCallbacks)
+	}
+
+	@Test
+	fun `run kin2 happy flow to completion`() {
```
yosriz, line 75: what about testing kin3 flow? can be very useful as you just change the logic there

yosriz (approved):
👍  some minor comments

core/src/test/kotlin/com/kin/ecosystem/core/data/order/ExternalEarnOrderCallTest.kt
OrenZak: yes, I will but not on this task.

core/src/test/kotlin/com/kin/ecosystem/core/data/order/ExternalEarnOrderCallTest.kt
OrenZak: 👍

PR #319, opened 2019-06-26
Gifting API

OrenZak:
#### Main purpose:
 Allow the developer to open a gifting dialog and use it as a p2p feature.
#### Technical description:
- Added dependency on `kin-appreciation-module-options-menu-android`
- Created `GiftingManager` and `GiftingManagerImpl` to support the whole feature.
- Create `JwtProvide` interface so the DS will need to implement and pass it as a param to our new API. 
- Added new API call to `Kin.java` -> `Kin.getGiftingManager(JwtProvider)`
#### Tests:
- Create to `GiftingManagerImpl`
#### Documentation:
Will be added on `partners-docs` repo

### DIFF
--- a/app/src/main/java/com/ecosystem/kin/app/KinJwtProvider.kt
+++ b/app/src/main/java/com/ecosystem/kin/app/KinJwtProvider.kt
@@ -1,0 +1,10 @@
+package com.ecosystem.kin.app
+
+import com.kin.ecosystem.JwtProvider
+
+class KinJwtProvider(private val appId: String,
+                     private val userId: String,
+                     private val deviceId: String) : JwtProvider {
+
+	override fun getPayToUserJwt(recipientUserID: String, amount: Double): String? = JwtUtil.generateGiftOfferExampleJWT(appId, userId, deviceId, recipientUserID, amount.toInt())
+}
--- a/app/src/main/java/com/ecosystem/kin/app/main/MainActivity.java
+++ b/app/src/main/java/com/ecosystem/kin/app/main/MainActivity.java
@@ -25,6 +25,7 @@ import android.widget.TextView;
 import android.widget.Toast;
 import com.ecosystem.kin.app.BuildConfig;
 import com.ecosystem.kin.app.JwtUtil;
+import com.ecosystem.kin.app.KinJwtProvider;
 import com.ecosystem.kin.app.R;
 import com.ecosystem.kin.app.login.LoginActivity;
 import com.ecosystem.kin.app.model.SignInRepo;
@@ -33,6 +34,7 @@ import com.kin.ecosystem.Kin;
 import com.kin.ecosystem.common.KinCallback;
 import com.kin.ecosystem.common.NativeOfferClickEvent;
 import com.kin.ecosystem.common.Observer;
+import com.kin.ecosystem.common.Subscription;
 import com.kin.ecosystem.common.exception.ClientException;
 import com.kin.ecosystem.common.exception.KinEcosystemException;
 import com.kin.ecosystem.common.exception.ServiceException;
@@ -42,13 +44,13 @@ import com.kin.ecosystem.common.model.NativeOffer;
 import com.kin.ecosystem.common.model.NativeSpendOfferBuilder;
 import com.kin.ecosystem.common.model.OrderConfirmation;
 import com.kin.ecosystem.common.model.UserStats;
+import com.kin.ecosystem.gifting.GiftingManager;
 import com.kin.ecosystem.recovery.BackupAndRestore;
 import com.kin.ecosystem.recovery.BackupAndRestoreCallback;
 import com.kin.ecosystem.recovery.exception.BackupAndRestoreException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Random;
 
 
 public class MainActivity extends AppCompatActivity {
@@ -81,6 +83,8 @@ public class MainActivity extends AppCompatActivity {
 	private NativeOffer nativeEarnOffer;
 
 	private BackupAndRestore backupAndRestore;
+	private GiftingManager giftingManager;
+	private Subscription<OrderConfirmation> giftingOrderConfirmation;
 
 	private NativeOffer getNativeSpendOffer() {
 		return new NativeSpendOfferBuilder(String.valueOf(getRandomID()))
@@ -107,6 +111,9 @@ public class MainActivity extends AppCompatActivity {
 	protected void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		setContentView(R.layout.activity_main);
+
+		userID = SignInRepo.getUserId(getApplicationContext());
+		deviceID = SignInRepo.getDeviceId(getApplicationContext());
 
 		try {
 			backupAndRestore = Kin.getBackupAndRestoreManager(this);
@@ -135,8 +142,6 @@ public class MainActivity extends AppCompatActivity {
 			e.printStackTrace();
 		}
 
-		userID = SignInRepo.getUserId(getApplicationContext());
-		deviceID = SignInRepo.getDeviceId(getApplicationContext());
 		containerLayout = findViewById(R.id.container);
 		balanceView = findViewById(R.id.get_balance);
 		nativeSpendTextView = findViewById(R.id.native_spend_button);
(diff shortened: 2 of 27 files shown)


app/src/main/java/com/ecosystem/kin/app/KinJwtProvider.kt
```diff
@@ -0,0 +1,10 @@
+package com.ecosystem.kin.app
+
+import com.kin.ecosystem.JwtProvider
+
+class KinJwtProvider(private val appId: String,
+                     private val userId: String,
+                     private val deviceId: String) : JwtProvider {
+
```
drEfrati, line 8: As far as I know, you can omit the constructor keyword from the primary constructor if it does not have any annotations or visibility modifiers (public, private or protected).
So how does this work for you?
Is it works because the only thing you are doing in the constructor is an assignment, then you could use the Primary Constructor with a private Property?

app/src/main/java/com/ecosystem/kin/app/main/MainActivity.java
```diff
@@ -584,30 +589,35 @@ private void showPayToUserDialog(final View v) {
 		payToUserDialog.setOnDismissListener(new OnDismissListener() {
 			@Override
 			public void onDismiss(DialogInterface dialog) {
```
drEfrati, line 591: even though it is only the sample app I think you should extract some code here to methods, also for readability.
I also bearly can see what is going on(attach screenshot)
Capture

app/src/main/java/com/ecosystem/kin/app/main/PayToUserDialog.kt
```diff
@@ -3,24 +3,51 @@ package com.ecosystem.kin.app.main
...
+    var recipientId: String? = null
                 private set
 
+    var openGifting = false
+        private set
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.pay_to_user_dialog)
         pay_button.setOnClickListener {
```
drEfrati, line 21: Because of this new kotlin feature, I think we should change our IDs to camel case so it would be written like this, for example, userIdEditText.text and not user_id_edit_text.text which is ugly

https://alexdunn.org/2018/03/14/android-kotlin-basics-camel-case-or-underscore-notation/

core/src/main/java/com/kin/ecosystem/core/bi/events/GiftingButtonTapped.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.Map;
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+
+
+/**
+ * User click on gifting dialog button 
+ * 
+ */
+public class GiftingButtonTapped implements Event {
```
drEfrati, line 18: It is a new class so why not in kotlin?
Or is it a generated class?

sdk/build.gradle
```diff
@@ -33,6 +34,9 @@ dependencies {
     implementation devDependencies.constraintLayout
     implementation devDependencies.picasso
     implementation devDependencies.kotlin
+    implementation ('org.kin.ecosystem.appreciation.options.menu:ui:1.0.0') {
+        transitive = false
```
drEfrati, line 38: why do you need it?

sdk/src/main/java/com/kin/ecosystem/gifting/GiftingManagerImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+internal class GiftingManagerImpl(private val jwtProvider: JwtProvider,
+                                  private val blockchainSource: BlockchainSource,
+                                  private val orderDataSource: OrderDataSource,
+                                  private val eventLogger: EventLogger,
+                                  private val configuration: Configuration) : GiftingManager, EventsListener {
+
+	private val orderConfirmation: ObservableData<OrderConfirmation> = ObservableData.create()
+	private var currentRecipientUserID: String? = null
+	private var isShowing = false
+	private val mainHandler = object : Handler(Looper.getMainLooper()) {
```
drEfrati, line 35: Just checking if you are sure you want to do it in the UI thread

sdk/src/main/java/com/kin/ecosystem/gifting/SendGiftCall.kt
```diff
@@ -0,0 +1,52 @@
...
+import android.os.Message
+import com.kin.ecosystem.JwtProvider
+import com.kin.ecosystem.common.KinCallback
+import com.kin.ecosystem.common.exception.ClientException
+import com.kin.ecosystem.common.exception.KinEcosystemException
+import com.kin.ecosystem.common.model.OrderConfirmation
+import com.kin.ecosystem.core.data.order.OrderDataSource
+import com.kin.ecosystem.core.util.ErrorUtil
+
+internal class SendGiftCall(private val handler: Handler,
```
drEfrati, line 13: Just beware of memory leaks

sdk/src/test/kotlin/com/kin/ecosystem/gifting/GiftingManagerImplTest.kt
```diff
@@ -0,0 +1,143 @@
...
+import org.junit.runner.RunWith
+import org.kin.ecosystem.appreciation.options.menu.ui.CloseType
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import org.robolectric.annotation.Config
+import java.math.BigDecimal
+
+@Config(manifest = Config.NONE)
+@RunWith(RobolectricTestRunner::class)
+class GiftingManagerImplTest: BaseTestClass() {
```
drEfrati, line 31: Are you sure you covered all cases?

drEfrati (approved):
Looks good.
I added some questions.
Just let sit for a couple of minutes to see the App itself on phone and that's it

app/src/main/java/com/ecosystem/kin/app/main/PayToUserDialog.kt
OrenZak: not agree -> `pay_button` is a generated value from kotlin-extensions (from the layout) and on the view layout its bad practice to use with camel case.

core/src/main/java/com/kin/ecosystem/core/bi/events/GiftingButtonTapped.java
OrenZak: you can see on the top comment -> // Augmented by script

sdk/build.gradle
OrenZak: because `options-menu` repo is using the `base` module from here and this resolves duplications.

sdk/src/main/java/com/kin/ecosystem/gifting/GiftingManagerImpl.kt
OrenZak: yup posting a callback on the main thread

sdk/src/test/kotlin/com/kin/ecosystem/gifting/GiftingManagerImplTest.kt
OrenZak: hope so yes seems 100% coverage.

sdk/src/main/java/com/kin/ecosystem/gifting/SendGiftCall.kt
OrenZak: 👍

core/src/main/java/com/kin/ecosystem/core/bi/events/GiftingButtonTapped.java
OrenZak: generated

app/src/main/java/com/ecosystem/kin/app/main/MainActivity.java
OrenZak: 👍

app/src/main/java/com/ecosystem/kin/app/KinJwtProvider.kt
OrenZak: Not sure what are you saying here.
The `private` is just saying to kotlin not to create a getter and setter for this property.

PR #322, opened 2019-07-02
not enough kin page

OrenZak:
#### Main purpose:
Show not enough page
#### Technical description:
- Created `NotEnoughKinFragment` as the view
- Added more `Screen.Id` -> `NOT_ENOUGH_KIN`
- Handled animations of content sceen height
#### Tests:
- Added `NotEnoughKinPresenterTest`


sdk/src/main/java/com/kin/ecosystem/gifting/GiftingManagerImpl.kt
```diff
@@ -55,18 +62,32 @@ internal class GiftingManagerImpl(private val jwtProvider: JwtProvider,
 
 	override fun addOrderConfirmationObserver(observer: Observer<OrderConfirmation>) = orderConfirmation.subscribe(observer)!!
 
-	override fun showDialog(context: Context, recipientUserID: String) {
+	override fun showDialog(activity: Activity, recipientUserID: String) {
```
drEfrati, line 65: If you need the application context you can use that as the parameter and not an activity.
You can also check if it is the application context and if not then get it from the context.

drEfrati (approved):
Seems legit to me.
Maybe to get the full picture it is better that i will also see this UI in the App itself.

sdk/src/main/java/com/kin/ecosystem/gifting/GiftingManagerImpl.kt
OrenZak: It used to override pending transaction from activity to another so I can't just use context, I need the activity ref.

PR #331, opened 2019-07-15
Kinno 642 reciever support transaction history

shayb:
#### Main purpose:
add ability to receiver to show transaction history
#### Technical description:
(What was changed/added)
#### Dilemmas you faced with? 
(Not a must)
#### Tests:
(Main tests you added/changed)
#### Documentation:
(Source/readme.md)

### DIFF
--- a/core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
+++ b/core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
@@ -146,8 +146,8 @@ public class AuthRepository implements AuthDataSource {
 	}
 
     @Override
-    public boolean isCurrentAuthTokenExpired() {
-		if(localData.isLoggedIn()) {
+    public boolean isCurrentAuthTokenExpired() throws ClientException {
+		if(!localData.isLoggedIn()) {
 			if (cachedAuthToken != null) {
 				return isAuthTokenExpired(cachedAuthToken);
 			}
@@ -156,7 +156,7 @@ public class AuthRepository implements AuthDataSource {
 				return isAuthTokenExpired(token);
 			}
 		}
-        return false;
+        throw ErrorUtil.getClientException(ClientException.ACCOUNT_NOT_LOGGED_IN, null);
     }
 
 	@Override
--- a/core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
+++ b/core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
@@ -151,8 +151,9 @@ public class AuthRepository implements AuthDataSource {
 			if (cachedAuthToken != null) {
 				return isAuthTokenExpired(cachedAuthToken);
 			}
-			if (localData.getAuthTokenSync() != null) {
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
```diff
@@ -147,11 +147,14 @@ public AuthToken getAuthTokenSync() {
...
+		if(localData.isLoggedIn()) {
+			if (cachedAuthToken != null) {
+				return isAuthTokenExpired(cachedAuthToken);
+			}
+			final AuthToken token = localData.getAuthTokenSync();
+			if (token != null) {
+				return isAuthTokenExpired(token);
+			}
 		}
         return false;
```
shayb: maybe should be true ?

sdk/src/main/java/com/kin/ecosystem/transfer/view/AccountInfoActivity.java
```diff
@@ -40,11 +41,29 @@ public String getData() throws AccountInfoException {
...
+                    IncomingTransfer payload = new IncomingTransfer().appId(senderAppId).memo(memo).title(title).description(description).walletAddress("");
+                    if (orderRepository != null) {
+                        try {
+                            orderRepository.createIncomingTransferOrderSync(payload);
+                        } catch (KinEcosystemException e) {
+                            e.printStackTrace();
+                        }
+                    }
+                }
+            });
```
shayb: Do it async from the repo itself.

OrenZak (approved):
Please check after the changes 👍

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
shayb: @OrenZak  no this is correct - only check validity of the token when user is logged in

OrenZak: but returning `false` at the end means that the authToken is valid and not expired but the user is not logged in at all so it's not make sense.

sdk/src/main/java/com/kin/ecosystem/transfer/view/AccountInfoActivity.java
```diff
@@ -40,11 +41,29 @@ public String getData() throws AccountInfoException {
...
+                @Override
+                public void run() {
+                    String title = "Received Kin";
+                    String description = "From " + senderAppName;
+                    final OrderRepository orderRepository = OrderRepository.getInstance();
+                    IncomingTransfer payload = new IncomingTransfer().appId(senderAppId).memo(memo).title(title).description(description).walletAddress("");
+                    if (orderRepository != null) {
+                        try {
+                            orderRepository.createIncomingTransferOrderSync(payload);
+                        } catch (KinEcosystemException e) {
```
OrenZak: why it still sync?

core/src/main/java/com/kin/ecosystem/core/data/auth/AuthRepository.java
shayb: i will throw an exp

PR #332, opened 2019-07-16
ECO 1504/Disable RTL support

AliAsadi:
#### Main purpose:
In case a phone's language is set to a language that is read from right to left our UI appears flipped and broken

#### Technical description:
* disable rtl support (Marketplace/ OrderHistory/ Settings)


OrenZak (approved):
🥇 Great :)

PR #337, opened 2019-08-13
Clear cache on logout

### LINKED ISSUE #310: Question about the getUserLoginState, setJWT and logout methods
Hi @OrenZak , I have a question about the getUserLoginState, setJWT and logout methods

In Kin.login(JWT) method, you use AuthRepository.getInstance().getUserLoginState(JWT) to check the loginState. If the input user_id or device_id in jwt are different with the cached data, the logout method will be invoked. 

But the cached data will not be clear if Kin SDK is not logged in before because the result of isAccountLoggedIn.compareAndSet(true, false) is false in logout method.

For example, user use user_id_1 to get jwt_1 and auth_token_1 and login Kin Successfully, then he logout app without call Kin#logout(developers forget to call Kin.logout when logout app's account) and kill app. In next time he use user_id_2 to try to login, but SDK will use user_id_1's authToken to try log in because user_id_1's authToken is not expired.

AliAsadi:
#### Main purpose:
If userA call `AuthRepository.getInstance().getAccountInfo` successfully, his token info will be written into cache. But if `AuthRepository.getInstance().updateWalletAddress` or migrate methods get failure callback, the `sendLoginFailed` method is called and `isAccountLoggedIn` variable is false. When userA want to use another app account, the `Kin.logOut` method is invoked. But userA’s token info in cache is not cleared because of `isAccountLoggedIn` is false. Another app account will always use userA’s token to interactive with `api.kinmarketplace.com`. For the users who have the issue, they cannot use Kin anymore unless the previous account’s token is expired or they clear app data.

Closes #310 

#### Technical description:
* cleare cache always when logout

### DIFF
--- a/sdk/src/main/java/com/kin/ecosystem/Kin.java
+++ b/sdk/src/main/java/com/kin/ecosystem/Kin.java
@@ -424,12 +424,10 @@ public class Kin {
 	 */
 	public static void logout() throws ClientException {
 		checkInstanceNotNull();
-		if (isAccountLoggedIn.compareAndSet(true, false)) {
-			eventLogger.send(UserLogoutRequested.create());
-			Logger.log(new Log().withTag("Kin.java").text("logout").put("isAccountLoggedIn", isAccountLoggedIn));
-			AuthRepository.getInstance().logout();
-			clearCachedData();
-		}
+		eventLogger.send(UserLogoutRequested.create());
+		Logger.log(new Log().withTag("Kin.java").text("logout").put("isAccountLoggedIn", isAccountLoggedIn));
+		AuthRepository.getInstance().logout();
+		clearCachedData();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/com/kin/ecosystem/Kin.java
```diff
@@ -424,12 +424,10 @@ private static void checkInstanceNotNull() throws ClientException {
 	 */
 	public static void logout() throws ClientException {
 		checkInstanceNotNull();
-		if (isAccountLoggedIn.compareAndSet(true, false)) {
```
OrenZak, line 427: you should keep `isAccountLoggedIn.compareAndSet(true, false)`.

OrenZak (approved):
see my comments 💯

sdk/src/main/java/com/kin/ecosystem/Kin.java
AliAsadi: Thanks, Updated.

PR #338, opened 2019-08-14
Navigation to login screen and connect with a different account

AliAsadi:
#### Main purpose:
To be able to go back to the login activity and connect with a different account.

#### Technical description:
* keep login activity in the stack


OrenZak (approved):
👍

PR #339, opened 2019-08-15
Crash with Proguard/R8

AliAsadi:
#### Main purpose:

Fix crash happening when building release with proguard 

#### Technical description:
* update migration version


OrenZak (approved):
🥇

PR #340, opened 2019-09-04
Fix spend process (Transaction Timeout)

AliAsadi:
#### Main purpose:
* Fix spend process 
* To be able to send more than a single transaction using the purchase method
* avoid transaction timeout exception

Reproduce:
Send 20 transactions at the same time.

Expected:
20 transaction success.

Actual:
1 - success
19 - failed (Transaction Timeout)

#### Technical description:
* add queue to manage transaction


yosriz (approved):
LGTM, would add sync or make `isExternalSpendInProcess` volatile for thread safety

